PR #11, opened 2017-05-30
Correct populate catalog from files

laurianed:


src/main/java/org/ow2/proactive/catalog/rest/util/CatalogObjectJSONParser.java
```diff
@@ -0,0 +1,95 @@
...
+
+
+/**
+ * This class parses a JSON describing a catalog object
+ * @author ActiveEon Team
+ * @since 4/25/2017
+ */
+public final class CatalogObjectJSONParser {
+
+    private static ObjectMapper MAPPER = new ObjectMapper();
```
ShatalovYaroslav: maybe final?

src/main/java/org/ow2/proactive/catalog/rest/util/CatalogObjectJSONParser.java
```diff
@@ -0,0 +1,95 @@
...
+    public static class CatalogObjectData {
+
+        private String name;
+
+        private String kind;
+
+        private String commitMessage;
+
+        private String objectFileName;
+
```
ShatalovYaroslav: do you think we can also add contentType? Or we can do it latter with List<KeyValueMetadata>.

src/main/java/org/ow2/proactive/catalog/rest/util/parser/CatalogObjectParserFactory.java
```diff
@@ -45,7 +45,7 @@ public static CatalogObjectParserFactory get() {
         return factory;
     }
 
-    public CatalogObjectParser getParser(String type) {
+    public CatalogObjectParserInterface getParser(String type) {
         if ("workflow".equals(type))
```
ShatalovYaroslav: I think we should add enum with all kinds we support. So we will know

laurianed: I don't think because it can be anything, and if not found you use the default parser that does nothing

ShatalovYaroslav: ok

PR #22, opened 2017-06-09
refactoring and code qulity improvement

ShatalovYaroslav:


src/main/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectRevisionController.java
```diff
@@ -108,15 +108,15 @@ public CatalogObjectMetadata create(@PathVariable Long bucketId, @PathVariable L
...
         return catalogObjectRevisionService.listCatalogObjectRevisions(bucketId, objectId, pageable, assembler);
     }
 
-    @ApiOperation(value = "Delete a catalog object's revision", notes = "If the revisionId references the latest revision, it is deleted and the catalog object then points to the previous revision. If the revisionId doesn't references the latest revision, it is simply deleted without any impact on the current catalog object. Returns the deleted CatalogObjectRevision metadata.")
-    @ApiResponses(value = @ApiResponse(code = 404, message = "Bucket or catalog object not found"))
+    @ApiOperation(value = "Delete a catalog object's revision", notes = "If the revisionId references the latest revision, it is deleted and the catalog object then points to the previous revision. If the revisionId doesn't references the latest revision, it is simply deleted without any impact on the current catalog object. Returns the deleted CatalogObjectRevisionEntity metadata.")
+    @ApiResponses(value = @ApiResponse(code = 404, message = "BucketEntity or catalog object not found"))
```
yinan-liu: for the message, i think we should better keep Bucket, not BucketEntity.

src/main/java/org/ow2/proactive/catalog/rest/entity/CatalogObjectRevisionEntity.java
```diff
@@ -215,14 +215,14 @@ public void setRawObject(byte[] rawObject) {
 
     @Override
     public String toString() {
-        return "CatalogObjectRevision{" + "commitId=" + commitId + ", kind='" + kind + '\'' + ", commitMessage='" +
-               commitMessage + '\'' + ", commitDate=" + commitDate + ", name='" + name + '\'' + ", bucketId=" +
-               bucketId + ", contentType='" + contentType + '\'' + ", keyValueMetadataList=" + keyValueMetadataList +
-               '}';
+        return "CatalogObjectRevisionEntity{" + "commitId=" + commitId + ", kind='" + kind + '\'' +
```
yinan-liu: if we use Objects.toString,  or apache common ToStringBuilder library to construct toString method, it will be much easier

PR #24, opened 2017-06-12
Add name and project name to workflow catalog object metadatas

laurianed:


src/main/java/org/ow2/proactive/catalog/rest/util/parser/WorkflowParser.java
```diff
@@ -51,6 +51,16 @@
...
 
+    private static final String JOB_NAME_KEY = "name";
+
+    private static final String PROJECT_NAME_KEY = "project_name";
+
+    private static final String JOB_AND_PROJECT_LABEL = "job_information";
+
+    private static final String ATTRIBUTE_JOB_NAME = "name";
+
+    private static final String ATTRIBUTE_JOB_PROJECT_NAME = "projectName";
```
marcocast: why not with the underscore?

laurianed: it's the name of the attribute into XML job file

PR #29, opened 2017-06-28
It is possible to filter buckets according to te kind of objects they…

laurianed:
… contain

### DIFF
--- a/src/main/java/org/ow2/proactive/catalog/service/BucketService.java
+++ b/src/main/java/org/ow2/proactive/catalog/service/BucketService.java
@@ -173,11 +173,13 @@ public class BucketService {
         return new BucketMetadata(bucket);
     }
 
-    public List<BucketMetadata> listBuckets(Optional<String> ownerName) {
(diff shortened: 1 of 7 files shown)


src/main/java/org/ow2/proactive/catalog/service/BucketService.java
```diff
@@ -173,11 +173,13 @@ public BucketMetadata getBucketMetadata(long id) {
         return new BucketMetadata(bucket);
     }
 
-    public List<BucketMetadata> listBuckets(Optional<String> ownerName) {
+    public List<BucketMetadata> listBuckets(Optional<String> ownerName, Optional<String> kind) {
         List<BucketEntity> entities;
 
         if (ownerName.isPresent()) {
```
yinan-liu: what about if both owner and kind are present ?

laurianed: I didn't see the use of filtering according to both fields. So if both are present only the user will be taken into account. Do you think i should change this?

PR #34, opened 2017-07-03
delete the specific empty bucket

ShatalovYaroslav:


yinan-liu:
code is not spotlessed, that's why jenkins job failed

src/main/java/org/ow2/proactive/catalog/service/BucketService.java
```diff
@@ -193,4 +194,18 @@ public void cleanAll() {
         bucketRepository.deleteAll();
     }
 
+    public BucketMetadata deleteEmptyBucket(long bucketId) {
+        BucketEntity bucket = bucketRepository.findBucketForUpdate(bucketId);
+
+        if (bucket == null) {
```
yinan-liu: i am not very sure, but when the bucketId does not exist here, the query will throw an exception instead of null maybe. Can you give just a manual test please. let me know the result please

ShatalovYaroslav: I've added the integration test for this case.

PR #38, opened 2017-07-17
graphql endpoint

yinan-liu:
new rest endpoint using graphql to fetch catalogobjects


src/main/java/org/ow2/proactive/catalog/graphql/bean/argument/CatalogObjectRevisionWhereArgs.java
```diff
@@ -35,13 +36,13 @@
...
 @EqualsAndHashCode(callSuper = true)
-public class RevisionWhereArgs extends BaseWhereArgs<Long> {
+public class CatalogObjectRevisionWhereArgs extends WhereArgs<Long> {
 
     private final Boolean isLatest;
 
-    public RevisionWhereArgs(Long eq, Long ne, Long gt, Long gte, Long lt, Long lte, Long[] in, Long[] notIn,
-            Boolean isNull, Boolean isLatest) {
-        super(eq, ne, gt, gte, lt, lte, in, notIn, isNull);
+    @Builder
```
sophiesongge: Could you please add a comment before the constructor of WhereArgs.java to explain what does this eq, ne, gt ... etc.. mean

src/main/java/org/ow2/proactive/catalog/graphql/bean/argument/WhereArgs.java
```diff
@@ -23,18 +23,39 @@
...
+    protected T ne;
+
+    protected T gt;
+
+    protected T gte;
+
+    protected T lt;
+
+    protected T lte;
+
```
sophiesongge: Could you please add a comment before the constructor of WhereArgs.java to explain what does this eq, ne, gt ... etc.. mean

yinan-liu: sure, it's jpa standard operation field name. i will put the comments on them

yinan-liu: done

src/main/java/org/ow2/proactive/catalog/graphql/handler/catalogobject/CatalogObjectAndOrGroupFilterHandler.java
```diff
@@ -0,0 +1,237 @@
...
+
+        // post-order traverse tree algorithm
+        Deque<List<CatalogObjectWhereArgs>> stack = new LinkedList<>();
+        stack.push(andOrArgs);
+
+        List<CatalogObjectWhereArgsTreeNode> ret = new ArrayList<>();
+
+        while (!stack.isEmpty()) {
+            List<CatalogObjectWhereArgs> top = stack.pop();
+            if (top != null) {
```
sophiesongge: Here, I prefer not to judge the top, but to judge top.left and top.right, which means that when top.left != null, we push top.left to the stack, then when top.right != null, we push top.right to the stack. To do so, we do not need to push so many null to the stack.

src/main/java/org/ow2/proactive/catalog/repository/specification/generic/ComparableSpecification.java
```diff
@@ -0,0 +1,69 @@
...
+            Join<CatalogObjectRevisionEntity, CatalogObjectEntity> catalogObjectJoin,
+            Join<CatalogObjectRevisionEntity, KeyValueMetadataEntity> metadataJoin) {
+        super(entityMetaModelEnum, operations, value, catalogObjectJoin, metadataJoin);
+    }
+
+    @Override
+    protected Predicate buildPredicate(Root<CatalogObjectRevisionEntity> root, CriteriaQuery<?> query,
+            CriteriaBuilder cb) {
+        switch (operations) {
+            case GT:
```
laurianed: Is it normal that it's the same action for (`GT` and `LT`) and (`GTE` and `LTE`) ? If yes, than you could replace by
```
case GT:
case LT:
return cb.equal(catalogObjectJoin.<R> get(entityMetaModelEnum.getName()), value);
case GTE:
case LTE:
return cb.notEqual(catalogObjectJoin.<R> get(entityMetaModelEnum.getName()), value);
```

yinan-liu: good catch.  copy-paste error !

PR #40, opened 2017-07-19
Catalog objects can be exported in a ZIP archive

laurianed:


src/main/java/org/ow2/proactive/catalog/service/CatalogObjectService.java
```diff
@@ -154,6 +159,19 @@ private CatalogObjectRevisionEntity buildCatalogObjectRevisionEntity(String kind
         return buildMetadataWithLink(bucketId, result);
     }
 
+    public byte[] getCatalogObjectsAsZipArchive(Long bucketId, List<String> catalogObjectsNames) {
+        List<CatalogObjectRevisionEntity> revisions = new ArrayList<>();
+
+        for (String name : catalogObjectsNames) {
+            CatalogObjectRevisionEntity catalogObjectRevision = catalogObjectRevisionRepository.findDefaultCatalogObjectByNameInBucket(bucketId,
+                                                                                                                                       name);
+            if (catalogObjectRevision == null)
```
MaelAudren: Why returning null when an object is missing ? 
If a missing object is not a normal behavior it would be better to through an exception. Otherwise if it is not an issue I think that returning the a zip without the missing object would be a better solution.

src/main/java/org/ow2/proactive/catalog/util/ArchiveManagerHelper.java
```diff
@@ -0,0 +1,113 @@
...
+                                                                    .map(catalogObjectRevision -> new ByteSource(catalogObjectRevision.getCatalogObject()
+                                                                                                                                      .getId()
+                                                                                                                                      .getName(),
+                                                                                                                 catalogObjectRevision.getRawObject()));
+            ZipEntrySource[] sources = streamSources.toArray(size -> new ZipEntrySource[size]);
+            ZipUtil.pack(sources, byteArrayOutputStream);
+
+            return byteArrayOutputStream.toByteArray();
+        } catch (IOException ioe) {
+            throw new RuntimeException(ioe);
```
MaelAudren: see upper comment

src/main/java/org/ow2/proactive/catalog/util/ArchiveManagerHelper.java
```diff
@@ -0,0 +1,113 @@
...
+
+        List<byte[]> filesList = new ArrayList<>();
+        if (byteArrayArchive == null) {
+            return filesList;
+        }
+
+        try (ByteArrayInputStream byteArrayInputStream = new ByteArrayInputStream(byteArrayArchive)) {
+            ZipUtil.iterate(byteArrayInputStream, (in, zipEntry) -> process(in, zipEntry, filesList));
+        } catch (IOException ioe) {
+            throw new RuntimeException(ioe);
```
MaelAudren: see upper comment

src/main/java/org/ow2/proactive/catalog/util/ArchiveManagerHelper.java
```diff
@@ -0,0 +1,113 @@
...
+     */
+    private void process(InputStream in, ZipEntry zipEntry, List<byte[]> filesList) {
+        try (ByteArrayOutputStream outputStream = new ByteArrayOutputStream()) {
+            int data = 0;
+            while ((data = in.read()) != -1) {
+                outputStream.write(data);
+            }
+            filesList.add(outputStream.toByteArray());
+        } catch (IOException ioe) {
+            throw new RuntimeException(ioe);
```
MaelAudren: see upper comment

src/main/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectController.java
```diff
@@ -152,25 +158,57 @@ public CatalogObjectMetadataList create(@PathVariable Long bucketId,
...
+    @ApiResponses(value = { @ApiResponse(code = 404, message = "Bucket not found"),
+                            @ApiResponse(code = 206, message = "Missing object") })
+    @RequestMapping(method = GET)
     @ResponseBody
-    public List<CatalogObjectMetadata> list(@PathVariable Long bucketId,
-            @ApiParam(value = "Filter according to kind.") @RequestParam(required = false) Optional<String> kind) {
-        List<CatalogObjectMetadata> result;
+    public ResponseEntity<?> list(@PathVariable Long bucketId,
+            @ApiParam(value = "Filter according to kind.") @RequestParam(required = false) Optional<String> kind,
+            @ApiParam(value = "Get given list in an archive") @RequestParam(value = "name", required = false) Optional<List<String>> names,
```
tobwiens: @RequestParam(value = "name") and the variable name is names, is that intended?

src/main/java/org/ow2/proactive/catalog/util/ArchiveManagerHelper.java
```diff
@@ -0,0 +1,147 @@
...
+import org.springframework.stereotype.Component;
+import org.zeroturnaround.zip.ByteSource;
+import org.zeroturnaround.zip.ZipEntrySource;
+import org.zeroturnaround.zip.ZipUtil;
+
+
+@Component
+public class ArchiveManagerHelper {
+
+    public static class ZipArchiveContent {
```
tobwiens: You introduce a global state here. 
Why is this state necessary? 
Is there a another way of doing it?

src/main/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectController.java
laurianed: Yes because it's a list.... You use it as `url?name=name1&name=name2...`

src/main/java/org/ow2/proactive/catalog/util/ArchiveManagerHelper.java
laurianed: Yes i need it to send 2 information when the archive is created : the created archive, and whether the archive is complete

PR #42, opened 2017-07-24
Objects can be uploaded to the catalog in a ZIP archive

laurianed:


src/main/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectController.java
```diff
@@ -90,20 +90,33 @@
     @RequestMapping(consumes = { MediaType.MULTIPART_FORM_DATA_VALUE }, method = POST)
     @ResponseStatus(HttpStatus.CREATED)
     public CatalogObjectMetadataList create(@PathVariable Long bucketId,
-            @ApiParam(value = "Name of the object") @RequestParam String name,
+            @ApiParam(value = "Name of the object ; if no name is provided, then the file should be a ZIP archive containing objects to create into the catalog") @RequestParam(required = false) Optional<String> name,
```
tobwiens: Instead of should, I would use something else to better indicate failure in case no name is provided. 
Something like:

"Name of the object or empty when a ZIP archive is uploaded[. All objects inside the archive are stored inside the catalog.]"

src/main/java/org/ow2/proactive/catalog/service/CatalogObjectService.java
```diff
@@ -89,6 +91,32 @@ public CatalogObjectMetadata createCatalogObject(Long bucketId, String name, Str
                                         rawObject);
     }
 
+    public List<CatalogObjectMetadata> createCatalogObjects(Long bucketId, String kind, String commitMessage,
+            String contentType, byte[] zipArchive) {
+
+        List<FileNameAndContent> filesContainedInArchive = archiveManager.extractZIP(zipArchive);
+
+        if (filesContainedInArchive.isEmpty()) {
+            throw new UnprocessableEntityException("Malformed archive");
```
tobwiens: Do we have some documentation where we describe this error behavior?

src/main/java/org/ow2/proactive/catalog/util/ArchiveManagerHelper.java
```diff
@@ -133,15 +162,27 @@ public ZipArchiveContent compressZIP(List<CatalogObjectRevisionEntity> catalogOb
...
             }
-            filesList.add(outputStream.toByteArray());
+            file.setContent(outputStream.toByteArray());
+            return file;
         } catch (IOException ioe) {
             throw new RuntimeException(ioe);
         }
     }
+
+    private String getFileNameWithoutExtension(String fileName) {
```
tobwiens: You can use FilenameUtils.getBasename

src/main/java/org/ow2/proactive/catalog/util/ArchiveManagerHelper.java
```diff
@@ -133,15 +162,27 @@ public ZipArchiveContent compressZIP(List<CatalogObjectRevisionEntity> catalogOb
      * @param zipEntry entry
      * @param filesList list of files
      */
-    private void process(InputStream in, ZipEntry zipEntry, List<byte[]> filesList) {
+    private FileNameAndContent process(InputStream in, ZipEntry entry) {
         try (ByteArrayOutputStream outputStream = new ByteArrayOutputStream()) {
+            FileNameAndContent file = new FileNameAndContent();
+            file.setName(getFileNameWithoutExtension(entry.getName()));
```
tobwiens: Use FilenameUtils.getBasename

src/main/java/org/ow2/proactive/catalog/service/CatalogObjectService.java
laurianed: Yes `@ApiResponse(code = 422, message = "Invalid file content supplied")`

PR #43, opened 2017-08-02
A new method PUT enables to restore an old revision as the last revision

laurianed:

### DIFF
--- a/src/main/java/org/ow2/proactive/catalog/dto/CatalogObjectMetadata.java
+++ b/src/main/java/org/ow2/proactive/catalog/dto/CatalogObjectMetadata.java
@@ -50,6 +50,9 @@ import lombok.EqualsAndHashCode;
 @EqualsAndHashCode(callSuper = false)
 public class CatalogObjectMetadata extends ResourceSupport {
 
+    @JsonProperty("commit_time_raw")
+    protected final String commitTimeRaw;
+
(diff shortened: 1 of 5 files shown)


src/main/java/org/ow2/proactive/catalog/dto/CatalogObjectMetadata.java
```diff
@@ -50,6 +50,9 @@
 @EqualsAndHashCode(callSuper = false)
 public class CatalogObjectMetadata extends ResourceSupport {
 
+    @JsonProperty("commit_time_raw")
+    protected final String commitTimeRaw;
```
ShatalovYaroslav: So it means that we will have commit_time_raw and commit_time. From the beginning when object was posted these times will be the same. But when we restore revision we update commit_time. Right?
Was it the needed requirement to have both dates? Why not to use only commit_time?

laurianed: the current commit_time is the commit time but displayed in a human-understandable way. I created  commit_time_raw which is the same time, but as a long (it's the ID of the revision)

ShatalovYaroslav: ok

PR #45, opened 2017-08-03
Secure catalog

tobwiens:
- Add sessionId to all endpoints
- Make sessionId check optional and configurable
- List only buckets of a users group
- Add Scheduler Authentication Service (Not fully developed)
- Add simple groups filtering to mock authorization

### DIFF
--- a/src/integration-test/resources/application-test.properties
+++ b/src/integration-test/resources/application-test.properties
@@ -22,3 +22,9 @@ spring.main.banner_mode=off
 
 spring.datasource.username=root
 spring.datasource.password=
+
+# Using the Scheduler as Authentication Service
+pa.scheduler.rest.url=http://localhost:8080/rest
+
+# Optional catalog security features
+pa.catalog.security.required.sessionid=false
--- a/src/main/java/org/ow2/proactive/catalog/rest/controller/BucketController.java
+++ b/src/main/java/org/ow2/proactive/catalog/rest/controller/BucketController.java
@@ -30,16 +30,21 @@ import static org.springframework.web.bind.annotation.RequestMethod.GET;
 import static org.springframework.web.bind.annotation.RequestMethod.POST;
 
 import java.util.List;
-import java.util.Optional;
 
 import org.ow2.proactive.catalog.dto.BucketMetadata;
 import org.ow2.proactive.catalog.service.BucketService;
+import org.ow2.proactive.catalog.service.RestApiAccessService;
+import org.ow2.proactive.catalog.service.exception.AccessDeniedException;
 import org.ow2.proactive.catalog.service.exception.BucketAlreadyExistingException;
+import org.ow2.proactive.catalog.service.exception.NotAuthenticatedException;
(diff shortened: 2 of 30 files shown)


src/integration-test/resources/application-test.properties
```diff
@@ -22,3 +22,9 @@ spring.main.banner_mode=off
 
 spring.datasource.username=root
 spring.datasource.password=
+
+# Using the Scheduler as Authentication Service
+pa.scheduler.rest.url=http://localhost:8080/rest
```
marcocast: Can you double check with @paraita how he solved the url catalog on the studio and see if you can apply the same fix here?

src/main/java/org/ow2/proactive/catalog/rest/controller/BucketController.java
```diff
@@ -61,11 +66,27 @@
...
+    @ApiResponses(value = { @ApiResponse(code = 401, message = "User not authenticated"),
+                            @ApiResponse(code = 403, message = "Permission denied"), })
     @RequestMapping(method = POST)
     @ResponseStatus(HttpStatus.CREATED)
-    public BucketMetadata create(@RequestParam(value = "name", required = true) String bucketName,
-            @ApiParam(value = "The name of the user that will own the Bucket") @RequestParam(value = "owner", required = true) String ownerName) {
+    public BucketMetadata create(
+            @ApiParam(value = "sessionID", required = false) @RequestHeader(value = "sessionID", required = false) String sessionId,
+            @RequestParam(value = "name", required = true) String bucketName,
+            @ApiParam(value = "The name of the user that will own the Bucket", defaultValue = "group:" +
```
marcocast: group should be UPPERCASE

src/main/java/org/ow2/proactive/catalog/service/AuthorizationService.java
```diff
@@ -0,0 +1,56 @@
...
+
+
+/**
+ * @author ActiveEon Team
+ * @since 27/07/2017
+ */
+@Component
+public class AuthorizationService {
+
+    private static final String GROUP_PREFIX = "group:";
```
marcocast: UPPERCASE, and it can be public and all the other references can point to this public static string

src/main/resources/application.properties
```diff
@@ -44,3 +44,9 @@ spring.jpa.show-sql=false
 
 # Disable Spring banner
 spring.main.banner_mode=off
+
+# Using the Scheduler as Authentication Service
+pa.scheduler.rest.url=http://localhost:8080/rest
+
+# Optional catalog security features
+pa.catalog.security.required.sessionid=true
```
marcocast: this should be false by default

src/integration-test/resources/application-test.properties
paraita: If https is not served by Nginx (so if it's served by the release's Jetty) you'll need to modify this value. We cannot infer the value at runtime because we don't have a discovery mechanism that would allow us to know if it's http/https and the port.
Otherwise, it will work just fine (it's the case on our try* platforms).

PR #53, opened 2017-08-08
Fix configuration bug and GROUP: prefix

tobwiens:
- The rest url was wrongly configured in the config, so resteasy threw an IllegalArguments Exception
- The GROUP: prefix must be added to the groups which come from the scheduler API. Because the owner is saved inside the DB with GROUP:{group}.
- The catalog is now populated instead of public-objects with GROUP:public-objects

### DIFF
--- a/src/main/java/org/ow2/proactive/catalog/service/BucketService.java
+++ b/src/main/java/org/ow2/proactive/catalog/service/BucketService.java
@@ -67,7 +67,7 @@ import lombok.extern.log4j.Log4j2;
 @Transactional
 public class BucketService {
 
-    public static final String DEFAULT_BUCKET_OWNER = "public-objects";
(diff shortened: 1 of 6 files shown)


src/main/java/org/ow2/proactive/catalog/service/BucketService.java
```diff
@@ -67,7 +67,7 @@
 @Transactional
 public class BucketService {
 
-    public static final String DEFAULT_BUCKET_OWNER = "public-objects";
+    public static final String DEFAULT_BUCKET_OWNER = "GROUP:public-objects";
```
marcocast: This should be : 

public static final String DEFAULT_BUCKET_OWNER = AuthorizationService.GROUP + ":public-objects";

tobwiens: Fixed

PR #59, opened 2017-08-16
Add metadata to xml file

tobwiens:


src/main/java/org/ow2/proactive/catalog/repository/entity/KeyValueMetadataEntity.java
```diff
@@ -113,7 +109,7 @@ public boolean equals(Object o) {
     public int hashCode() {
         int result = super.hashCode();
         result = 31 * result + key.hashCode();
-        result = 31 * result + catalogObjectRevision.hashCode();
+        result = 31 * result + ((catalogObjectRevision == null) ? 0 : catalogObjectRevision.hashCode());
```
tobwiens: When one puts the KeyValueMetadataEntity into a hashMap it fails because catalogObjectRevision can be null.

src/test/java/org/ow2/proactive/catalog/service/WorkflowXmlManipulatorTest.java
```diff
@@ -0,0 +1,138 @@
...
+
+
+/**
+ * @author ActiveEon Team
+ * @since 16/08/2017
+ */
+@RunWith(MockitoJUnitRunner.class)
+public class WorkflowXmlManipulatorTest {
+
+    private final byte[] simpleWorkflowWithoutGenericInfo = ("<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n" + "<job\n" +
```
MaelAudren: Maybe it would be better to serialize a workflow file into a String instead of describing it entirely in the code.

src/main/java/org/ow2/proactive/catalog/repository/entity/KeyValueMetadataEntity.java
```diff
@@ -62,25 +62,26 @@
...
     @JoinColumns({ @JoinColumn(name = "CATALOGOBJECTREVISION", referencedColumnName = "ID") })
     protected CatalogObjectRevisionEntity catalogObjectRevision;
 
+    @GeneratedValue(generator = "uuid2")
+    @GenericGenerator(name = "uuid2", strategy = "uuid2")
+    @Column(name = "ID", columnDefinition = "BINARY(16)")
+    @Id
+    private UUID id;
+
     public KeyValueMetadataEntity(String key, String value, String label) {
```
marcocast: why we have a label parameter if we don t use it after?maybe it is better to remove it

src/main/java/org/ow2/proactive/catalog/service/KeyValueMetadataHelper.java
```diff
@@ -34,17 +37,35 @@
...
 /**
  * @author ActiveEon Team
  * @since 19/06/2017
  */
+@Component
 public class KeyValueMetadataHelper {
 
-    public static List<KeyValueMetadataEntity> extractKeyValuesFromRaw(String kind, byte[] rawObject) {
+    @SuppressWarnings("FieldCanBeLocal")
+    private static final String GROUP_KEY = "group";
```
marcocast: here is lower case and in the OwnerGroupStringHelper is upper case....is that correct?

tobwiens: Yes the lower case "group" is the cone which is inside the generic information. From what I remember that would be a lower case "group". 

Maybe we can change it to a more concise pattern? Do you have ideas?

marcocast: what if we use uppercase group all over the places? inside the generic information you can put anything you want and then you just need to match the key in the security file

src/main/java/org/ow2/proactive/catalog/repository/entity/KeyValueMetadataEntity.java
tobwiens: We use the label parameter, when we store these additional information (KeyValueMetadataEntity) inside the database.

I think that parameter is intended to allow the search by generic information or other job related information without reading the xml inside the database. 

Currently we use the key as an uniqueness constraint, which creates problems see this issue: https://github.com/ow2-proactive/catalog/issues/60 . In order to fix that, we could widen the uniqueness constraint to key and label. Which would allow us to have a generic information with the key "name" (because the workflow name is stored inside a KeyValueMetadataEntity with the key = "name" -> that violates the uniqueness constraint). 

But it is probably better to ask people who initially worked on the catalog and made those decisions.@yinan-liu can help us out on this question?

src/test/java/org/ow2/proactive/catalog/service/WorkflowXmlManipulatorTest.java
tobwiens: Yeah right it makes it messy, I will fix that.

src/main/java/org/ow2/proactive/catalog/repository/entity/KeyValueMetadataEntity.java
ShatalovYaroslav: @tobwiens This is good idea. I was already thinking about this.
If we add to @UniqueConstraint besides "CATALOGOBJECTREVISION", "KEY" 
also Label it will solve all problems.

PR #61, opened 2017-08-16
Add Label field to uniqueness constraint of KeyValueMetadataEntity

### LINKED ISSUE #60: There is an error when publish workflow with generic info with Property Name "name"
It's not possible to publish workflow to catalog if the wf conatins generic info with Property Name "name".
This is due to unique constraint of key field inside the metadata table in catalog service.
Exact exception is:
org.hsqldb.HsqlException: integrity constraint violation: unique constraint or index violation; UKKJ251Y7PARAV20FN5HJGAHBKG table: METADATA_KEY_VALUE

ShatalovYaroslav:
This fixes #60 issue.

### DIFF
--- a/src/main/java/org/ow2/proactive/catalog/repository/entity/KeyValueMetadataEntity.java
+++ b/src/main/java/org/ow2/proactive/catalog/repository/entity/KeyValueMetadataEntity.java
@@ -57,8 +57,9 @@ import lombok.NoArgsConstructor;
 @Entity
 @NoArgsConstructor
 @Table(name = "METADATA_KEY_VALUE", uniqueConstraints = @UniqueConstraint(columnNames = { "CATALOGOBJECTREVISION",
-                                                                                          "KEY" }), indexes = { @Index(columnList = "KEY"),
-                                                                                                                @Index(columnList = "VALUE") })
+                                                                                          "KEY",
+                                                                                          "LABEL" }), indexes = { @Index(columnList = "KEY"),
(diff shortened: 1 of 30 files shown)


src/main/java/org/ow2/proactive/catalog/repository/entity/KeyValueMetadataEntity.java
```diff
@@ -57,8 +57,9 @@
 @Entity
 @NoArgsConstructor
 @Table(name = "METADATA_KEY_VALUE", uniqueConstraints = @UniqueConstraint(columnNames = { "CATALOGOBJECTREVISION",
-                                                                                          "KEY" }), indexes = { @Index(columnList = "KEY"),
-                                                                                                                @Index(columnList = "VALUE") })
+                                                                                          "KEY",
+                                                                                          "LABEL" }), indexes = { @Index(columnList = "KEY"),
+                                                                                                                  @Index(columnList = "VALUE") })
 @Inheritance(strategy = InheritanceType.SINGLE_TABLE)
 public class KeyValueMetadataEntity implements Serializable {
```
marcocast: should this be renamed to KeyValueLabelMetadataEntity ?

tobwiens:
@ShatalovYaroslav I just realized you renamed KeyValueLabelMetadataEntity but not the helper classes and the  methods. I see if I can combine that inside my current PR, otherwise I open an issue.

tobwiens:
I added the renaming here: https://github.com/ow2-proactive/catalog/pull/64

PR #66, opened 2017-08-21
Fix the file permissions and the url of the cloud automation services

paraita:
This is related to the changes on the cloud-automation portal to fix the visual glitches that were spotted on release 7.29.
The default behaviour of the cloud automation portal is now to use a generic image resource located on the cloud automation portal when the `pca.action.icon` parameter has an empty value.


paraita:
Thanks @MaelAudren for the comment !
2 reasons:

* the referenced icon was causing the catalog visual area of the Cloud Automation subview to act weird when icons are not square (we still have the issue as it can't be fixed without having to remove the fluidness of the page, which is required to display the portal properly on hi-resolution screens, see [here](https://bitbucket.org/activeeon/cloud-automation/commits/2c4079052b545158dc6f995135b359b492f5d3df))
* we (wrongly) coupled this Studio resource to a workflow so if this resource changes in the future on the Studio side, it will have side effects on the Cloud-Automation portal as well. The new behaviour is that when a PCA workflow doesn't have any value for the icon, the Cloud-Automation portal shall display a default icon (see my other PR [here](https://bitbucket.org/activeeon/cloud-automation/pull-requests/51/small-visual-fixes-for-7-29-maintenance/diff)). The icon is the same, except it has been cropped to square.

PR #67, opened 2017-08-23
Only add default owner to empty owner list

tobwiens:
Problem:
- listing all buckets for GROUP:admin lists all buckets for GROUP:public-objects and GROUP:admin. My expectation is that when I "search" for a specific owner, I get only buckets of that owner.

Solution:
- Add the default owner GROUP:public-objects only if the query did not specify any owner. So a query which has no owner specified will list all buckets visible for the user.


tobwiens:
@MaelAudren @paraita Thank you for your reviews. Some tests were failing, so I fixed it and squashed the changes and pushed.

Issue #57, opened 2017-08-11
Get /buckets response  with kind and owner arguments is not coherent


MaelAudren:
With the following configuration : 
bucket1(owner:bot) containing an object with the kind "worklfow"
bucket2(owner:toto) containing an object with the kind "workflow"
bucket3(owner:bot) not containing an object with the kind "worklfow"

Get /buckets?owner=bot&kind=worklfow 
return bucket1 and bucket3

The response is neither a logic AND or OR. It seems that if owner parameter is passed, kind parameter is ignored.

ShatalovYaroslav:
It's en expecting behavior.
From specification it was decided, that we will get all buckets, that satisfy the condition for owner bucket and kind of objects inside this bucket, but also returning empty buckets. Because for empty buckets, we don't no what kind of objects will be inside bucket.
Is bucket3 is empty in your case?

MaelAudren:
The request will return the same response if bucket3 is empty or if it contains workflows with a kind which is not "workflow" .

PR #68, opened 2017-08-23
Get /buckets response with kind and owner arguments is coherent now for insecure mode

ShatalovYaroslav:
fix issue #57

### DIFF
--- a/src/main/java/org/ow2/proactive/catalog/service/SchedulerUserAuthenticationService.java
+++ b/src/main/java/org/ow2/proactive/catalog/service/SchedulerUserAuthenticationService.java
@@ -50,9 +50,15 @@ public class SchedulerUserAuthenticationService {
     }
 
     public AuthenticatedUser authenticateBySessionId(String sessionId) throws NotAuthenticatedException {
-        UserData userData = this.schedulerRestClientCreator.getNewClientInitializedWithSchedulerRestUrl()
-                                                           .getScheduler()
-                                                           .getUserDataFromSessionId(sessionId);
+        UserData userData;
+        try {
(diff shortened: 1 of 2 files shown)


src/main/java/org/ow2/proactive/catalog/service/BucketService.java
```diff
@@ -188,8 +188,12 @@ public BucketMetadata getBucketMetadata(long id) {
 
     public List<BucketMetadata> listBuckets(String ownerName, String kind) {
         List<BucketEntity> entities;
+        List<String> owners = new ArrayList<>();
+        owners.add(ownerName);
 
-        if (!StringUtils.isEmpty(ownerName)) {
+        if (!StringUtils.isEmpty(ownerName) && !StringUtils.isEmpty(kind)) {
+            entities = bucketRepository.findByOwnerIsInContainingKind(owners, kind);
```
tobwiens: You can use ```Collections.singletonList(ownerName)``` instead of:  
  ```     
List<String> owners = new ArrayList<>();
       owners.add(ownerName);
```

ShatalovYaroslav: Thanks for useful comments.
In the end of the day the secure method listBuckets(List<String> owners, String kind) and insecure listBuckets(String ownerName, String kind) method will call the same DB method, but only in case when the owner and kind is not empty. It will be better for system tests.
I didn't change the behavior of insecure mode (just added check when both parameters are present), before it was just ignoring kind when the both kind and owner are presented.
This is clear shown in the issue: https://github.com/ow2-proactive/catalog/issues/57

PR #74, opened 2017-08-29
XML encode generic information

tobwiens:


tobwiens:
The correct solution to this is, IMHO, to change the generic information when we get a raw object and only on request of the user.

PR #80, opened 2017-09-01
spring dependency management plugin

yinan-liu:


yinan-liu:
yes, @gheon , that's the goal of this pull request

PR #81, opened 2017-09-01
remove population of catalog on startup from including to this project raw objects

ShatalovYaroslav:
This PR will remove all objects, stored inside catalog project. These objects were used to populate catalog on  startup of service. Now the catalog should be populated with objects from https://github.com/ow2-proactive/proactive-examples via Rest api only.


paraita:
Does it mean that if someone compiles the catalog from source it won't have any objects anymore ? If someone downloads the repo, he will have difficulties to understand the proper way to inject such objects because there no more examples. Is that ok ?

ShatalovYaroslav:
@paraita In fact, the proper way to inject objects is only via REST api, and not from the service itself.
In order to inject any object the user will need just upload any kind of object (binary, image, ...) and follow the documentation to post the object. I agree that the documentation should be better, at least in readme file. Nevertheless there is still some examples of objects in the resources of tests.

paraita:
@ShatalovYaroslav ah yes the tests sorry it makes sense thanks for your answer !

PR #83, opened 2017-09-07
Inside the workflow replace generic information only on job level and not on task level

ShatalovYaroslav:


src/test/java/org/ow2/proactive/catalog/service/WorkflowXmlManipulatorTest.java
```diff
@@ -82,31 +84,106 @@
...
+                                                                                   "    </task>\n" + "  </taskFlow>\n" +
+                                                                                   "     <description>\n" +
+                                                                                   "        <![CDATA[ Perform anomaly detection of an input image (only pedestrians are supposed to be present) ]]>\n" +
+                                                                                   "     </description>\n" +
+                                                                                   "    <genericInformation>" +
+                                                                                   "      <info name=\"first\" value=\"value1\"/>\n" +
+                                                                                   "      <info name=\"second\" value=\"value2\"/>\n" +
+                                                                                   "    </genericInformation>\n" +
+                                                                                   "</job>").getBytes();
+
```
gheon: Just a question: isn't it better way to embed this data as external (assest) files?

PR #89, opened 2017-09-28
No session id for public objects

tobwiens:
Allow access to GROUP:public-objects without valid sessionId.


tobwiens:
We change the load-examples scripts instead.

PR #92, opened 2017-10-23
Make the name of the bucket - unique

### LINKED ISSUE #87: The name of the bucket should be unique
In order to simplify the model and usage of the catalog, we can remove the actual unicity constraint that we have on the owner+bucket_name and put it on the bucket_name only instead.

ShatalovYaroslav:
Fix #87 issue


cdiop:
Oki for me

Issue #41, opened 2017-07-21
Cannot get the raw object with some private content type in swagger


MaelAudren:
With the swagger interface :
Upload it with the content-type "myApp" on /catalog/swagger-ui.html#!/catalog-object-controller/createUsingPOST_1 
=> the object is well uploaded
Get the raw object on /catalog/swagger-ui.html#!/catalog-object-controller/getRawUsingGET
=> it fails 

It seems that it fail only through the swagger interface. 
If the content type is "application/myApp", it works.

ShatalovYaroslav:
According to discussion in card: 
https://trello.com/c/g0Qh92SR/2423-generic-catalog-remove-the-mediatype-check

Just to summarize the possible solutions for content-type:
1) leave as it is now: not adding custom content type to header, close issue
2) add custom content type: avoid creation of MediaType java object and validation
3) try to create MediaType from content-type, if exception, then put in header binary type: application/octet-stream
The choice of the solution depends if we want to allow user to get his own content-type or not.

Final decision: Let's leave it as it is now. Thanks

PR #94, opened 2017-11-23
add content disposition header when request raw object

ShatalovYaroslav:
Also this functionality was extracted to external class

### DIFF
--- a/src/main/java/org/ow2/proactive/catalog/util/RawObjectResponseCreator.java
+++ b/src/main/java/org/ow2/proactive/catalog/util/RawObjectResponseCreator.java
@@ -1,0 +1,82 @@
+/*
+ * ProActive Parallel Suite(TM):
+ * The Open Source library for parallel and distributed
+ * Workflows & Scheduling, Orchestration, Cloud Automation
+ * and Big Data Analysis on Enterprise Grids & Clouds.
+ *
+ * Copyright (c) 2007 - 2017 ActiveEon
+ * Contact: contact@activeeon.com
+ *
+ * This library is free software: you can redistribute it and/or
+ * modify it under the terms of the GNU Affero General Public License
+ * as published by the Free Software Foundation: version 3 of
+ * the License.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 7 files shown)


src/main/java/org/ow2/proactive/catalog/util/RawObjectResponseCreator.java
```diff
@@ -0,0 +1,82 @@
...
+        String name = rawObject.getName();
+
+        byte[] bytes = rawObject.getRawObject();
+
+        ResponseEntity.BodyBuilder responseBodyBuilder = ResponseEntity.ok().contentLength(bytes.length);
+
+        try {
+            String contentDispositionFileName = name;
+
+            //add the .xml extension to contentDispositionFileName for workflow if the extension was not yet in name
```
MaelAudren: rephrase in "for the workflow if the extension was not in the variable 'name'"

src/test/java/org/ow2/proactive/catalog/util/RawObjectResponseCreatorTest.java
```diff
@@ -0,0 +1,130 @@
...
+                                                          "name",
+                                                          "object",
+                                                          "testContentType",
+                                                          1400343L,
+                                                          "commit message",
+                                                          Collections.emptyList(),
+                                                          new byte[0]);
+        ResponseEntity responseEntity = rawObjectResponseCreator.createRawObjectResponse(rawObject);
+        assertThat(responseEntity).isNotNull();
+        assertThat(responseEntity.getStatusCode()).isEqualTo(HttpStatus.OK);
```
MaelAudren: The wrong content type return an OK response ?

ShatalovYaroslav: exactly. This what we decided long time ago in this issue:
https://github.com/ow2-proactive/catalog/issues/41

src/main/java/org/ow2/proactive/catalog/util/RawObjectResponseCreator.java
ShatalovYaroslav: ok

PR #99, opened 2018-01-05
Change the return type of CatalogObjectController methods

MaelAudren:
Change the return type of CatalogObjectController method in order to fix issues with the swagger model. It fix issues that prevent the generation of a usable catalog client.
It solves the swagger model issue "Not a valid ...File definition"

### DIFF
--- a/src/main/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectController.java
+++ b/src/main/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectController.java
@@ -30,6 +30,7 @@ import static org.springframework.web.bind.annotation.RequestMethod.GET;
 import static org.springframework.web.bind.annotation.RequestMethod.POST;
 import static org.springframework.web.bind.annotation.RequestMethod.PUT;
 
+import java.io.File;
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
 import java.net.MalformedURLException;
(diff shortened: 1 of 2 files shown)


src/main/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectController.java
```diff
@@ -242,7 +240,7 @@ public CatalogObjectMetadataList create(
             } catch (IOException ioe) {
                 throw new RuntimeException(ioe);
             }
-            return ResponseEntity.status(status).build();
+            return new ResponseEntity<>(status);
```
marcocast: I prefer to use the builder .... ResponseEntity.status(status).build();

src/main/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectController.java
```diff
@@ -302,7 +300,7 @@ public CatalogObjectMetadataList create(
                                                                                        commitTimeRaw);
             return ResponseEntity.ok(catalogObjectMetadata);
         } catch (RevisionNotFoundException e) {
-            return ResponseEntity.notFound().build();
+            return new ResponseEntity(HttpStatus.NOT_FOUND);
```
marcocast: better with the builder no?

MaelAudren: No it makes the swagger model generation fail

src/main/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectController.java
MaelAudren: ResponseEntity.status(status).build(); return a Response<Void> which doesn't compile due to the returne Response<List<CatalogObjectMetadata>>

src/main/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectController.java
MaelAudren: same as upper response

PR #118, opened 2018-06-04
change a filter for getting list of objects by kind prefix

ShatalovYaroslav:


src/integration-test/java/org/ow2/proactive/catalog/service/CatalogObjectServiceIntegrationTest.java
```diff
@@ -134,8 +134,19 @@ public void testListCatalogObjectsByKindInBucket() {
                                                                                                    "object");
         assertThat(catalogObjects).hasSize(2);
 
-        catalogObjects = catalogObjectService.listCatalogObjectsByKind(bucket.getName(), "workflow");
+        catalogObjectService.createCatalogObject(bucket.getName(),
+                                                 "catalog4",
```
chloegugli: "catalog4" is normal or you meant "catalog"?

ShatalovYaroslav: I just follow the logic of creating previous objects in this test with name catalog1, catalog2

PR #135, opened 2018-08-08
improve kind's name validation

ShatalovYaroslav:


src/integration-test/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectControllerIntegrationTest.java
```diff
@@ -207,7 +207,7 @@ public void testGetAllKindsFromCatalog() throws JsonProcessingException {
         allKinds.add(workflowKind);
         given().when().get(kindsQuery).then().assertThat().statusCode(HttpStatus.SC_OK).body("", is(allKinds));
 
-        String newKindMy = "workflow/new_kind/my";
+        String newKindMy = "workflow/new_kind/mine";
```
Mykhailenko: then do newKindMine)

PR #141, opened 2018-08-22
add content Type validation

ShatalovYaroslav:


src/main/java/org/ow2/proactive/catalog/util/name/validator/KindAndContentTypeValidator.java
```diff
@@ -29,15 +29,14 @@
...
- A kind name must start and terminate with a letter or numbers. So kind name cannot start and terminate with a dash, space, dot or underscore.
- A kind names can be separated by slash symbol.
+ * According to this validator: the name can be between 3 and 63 characters long, and can contain only letters, numbers and set of next characters _. ;=+-
+ A name must start and terminate with a letter or numbers.  The names can be separated by slash symbol.
  */
 @Component
-public class KindNameValidator extends NameValidator {
-    protected static final String VALID_KIND_NAME_PATTERN = "^([a-zA-Z0-9][a-zA-Z0-9_\\. -]{1,61}[a-zA-Z0-9]\\/?)+$";
+public class KindAndContentTypeValidator extends NameValidator {
+    protected static final String VALID_KIND_NAME_PATTERN = "^([a-zA-Z0-9][a-zA-Z0-9_\\. ;=\\+\\-]{1,61}[a-zA-Z0-9]\\/?)+$";
```
zeineb: You forgot to update the authorized length of kind/content type (3 instead of 3, and 63 instead of 61)

PR #143, opened 2018-08-23
Pdf report table style

marcocast:


src/main/java/org/ow2/proactive/catalog/report/TableDataBuilder.java
```diff
@@ -87,15 +95,35 @@ public void buildTableData(Set<CatalogObjectMetadata> orderedObjectsPerBucket, B
...
+                            .map(metadata -> "<li><b>" + metadata.getKey() + "</b> = " + metadata.getValue() + "</li>")
+                            .collect(Collectors.joining(""));
+
+    }
+
+    private String getVariable(CatalogObjectMetadata catalogObject) {
+        return catalogObject.getMetadataList()
+                            .stream()
+                            .filter(metadata -> metadata.getLabel().equals("variable"))
+                            .map(metadata -> "<li><b>" + metadata.getKey() + "</b> = " + metadata.getValue() + "</li>")
```
MaelAudren: If you are modifying the variable to include html code then rename the method to indicate it please. This is not the behavior that I would expect for such method name.

src/main/java/org/ow2/proactive/catalog/report/TableDataBuilder.java
```diff
@@ -129,9 +163,10 @@ private String getProjectName(CatalogObjectMetadata catalogObject) {
     private String getDescription(CatalogObjectMetadata catalogObject) {
         return catalogObject.getMetadataList()
                             .stream()
-                            .filter(metadata -> metadata.getKey().equals("description"))
+                            .filter(metadata -> metadata.getLabel().equals("General") &&
+                                                metadata.getKey().equals("description"))
                             .map(metadata -> metadata.getValue())
-                            .map(description -> description.replaceAll("\n", "").replace("\r", "").replace("\t", ""))
+                            .map(description -> description.replaceAll("\n", " ").replace("\r", "").replace("\t", ""))
```
jrochas: Why do you need to use `replaceAll` for the description? It does the same thing as `replace` but takes a regex as first parameter. I think it can be handled the same as `\r` and `\t`.

MaelAudren: It would be good to change the method name here for the same reason as the upper comment.

src/main/java/org/ow2/proactive/catalog/report/TableDataBuilder.java
```diff
@@ -87,15 +95,35 @@ public void buildTableData(Set<CatalogObjectMetadata> orderedObjectsPerBucket, B
...
+    private String getGenericInfo(CatalogObjectMetadata catalogObject) {
+        return catalogObject.getMetadataList()
+                            .stream()
+                            .filter(metadata -> metadata.getLabel().equals("generic_information"))
+                            .map(metadata -> "<li><b>" + metadata.getKey() + "</b> = " + metadata.getValue() + "</li>")
+                            .collect(Collectors.joining(""));
+
+    }
+
+    private String getVariable(CatalogObjectMetadata catalogObject) {
```
jrochas: Maybe we can rename with plural: `getVariables`, if it applies. Same comment for `getGenericInfo`.

src/main/java/org/ow2/proactive/catalog/report/TableDataBuilder.java
```diff
@@ -118,6 +146,12 @@ private void createDataCell(Row<PDPage> row, float width, String catalogObjectDa
         bucketNameCell.setFillColor(java.awt.Color.decode(color));
     }
 
+    private void createKeyValueContentDataCell(Row<PDPage> row, float width, String catalogObjectData, int size) {
+        Cell<PDPage> cell2 = row.createCell(width, "<ul>" + catalogObjectData + "</ul>");
```
jrochas: Is there a reason for the name `cell2`?

src/main/java/org/ow2/proactive/catalog/report/TableDataBuilder.java
```diff
@@ -118,6 +146,12 @@ private void createDataCell(Row<PDPage> row, float width, String catalogObjectDa
         bucketNameCell.setFillColor(java.awt.Color.decode(color));
     }
 
+    private void createKeyValueContentDataCell(Row<PDPage> row, float width, String catalogObjectData, int size) {
```
jrochas: We can rename `size` into `fontSize` and create a constant for it for the two usages of this method.

marcocast:
All reviews have been applied. Thanks!!!

PR #153, opened 2018-10-30
add integration test for the report generation

MaelAudren:


src/main/java/org/ow2/proactive/catalog/report/HeadersBuilder.java
```diff
@@ -63,10 +67,14 @@
...
-        BufferedImage imageFile = ImageIO.read(url);
-        headerRow.createImageCell((100 / 12f) * 3, new Image(imageFile));
-
+        try {
+            URL url = new URL(schedulerUrl + ACTIVEEON_LOGO);
+            BufferedImage imageFile = ImageIO.read(url);
+            headerRow.createImageCell((100 / 12f) * 3, new Image(imageFile));
+        } catch (Exception e) {
+            headerRow.createCell((100 / 12f) * 3, "Activeeon", HorizontalAlignment.CENTER, VerticalAlignment.MIDDLE);
+            log.error("Impossible to create the image from the url: " + schedulerUrl + ACTIVEEON_LOGO);
```
jrochas: Can you chain the exception `e` in the log?

MaelAudren: I can do that.

PR #158, opened 2018-12-04
Add support for Oracle, MySQL, Postgresql and SqlServer databases

nebilbenmabrouk:


build.gradle
```diff
@@ -45,6 +45,8 @@ bootRepackage {
     enabled = false
 }
 
+project.ext.set("local", "")
```
fviale: similar remark as in scheduling api, local should not be forced but passed as a build parameter -Plocal

build.gradle
```diff
@@ -173,8 +175,8 @@ dependencies {
      */
     compile 'org.springframework.boot:spring-boot-starter-data-jpa'
 
-    compile 'org.hibernate:hibernate-entitymanager'
-    compile 'org.hibernate:hibernate-core'
+    compile('org.hibernate:hibernate-entitymanager:5.1.16.Final')
+    compile('org.hibernate:hibernate-core:5.1.16.Final')
```
fviale: same remark as in the other project, this should be managed by parent bom

build.gradle
```diff
@@ -204,6 +206,18 @@ dependencies {
     // library to build pdf reports
     compile 'com.github.dhorions:boxable'
 
+    //support for Oracle DB version 12.1.0.2
+    providedCompile group: 'com.oracle', name: 'ojdbc7', version: '12.1.0.2'
```
fviale: why not testCompile ?

src/integration-test/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectControllerIntegrationTest.java
```diff
@@ -168,729 +168,753 @@ public void testCreateWorkflowShouldReturnSavedWorkflow() {
...
-               .queryParam("commitMessage", "Import from archive")
-               .queryParam("objectContentType", MediaType.APPLICATION_XML.toString())
-               .multiPart(IntegrationTestUtil.getArchiveFile("workflow_0.xml"))
-               .when()
-               .post(CATALOG_OBJECTS_RESOURCE)
-               .then()
-               .assertThat()
-               .statusCode(HttpStatus.SC_UNPROCESSABLE_ENTITY);
-    }
+    /*
```
fviale: Commenting tests makes no sense, you can use @Ignore with an explanation.

At any rate, why these tests must be removed ? We reduce the testing span doing this, and our ability to spot regressions.

src/integration-test/resources/application-test.properties
```diff
@@ -4,6 +4,54 @@ logging.level.org.hibernate.SQL=off
...
+######################## HSQL DB #######################################################
+# The default settings are using hsqldb
+spring.datasource.driverClassName=org.hsqldb.jdbc.JDBCDriver
+spring.datasource.url=jdbc:hsqldb:mem:catalog
+spring.jpa.database-platform=org.hibernate.dialect.HSQLDialect
+
+spring.datasource.username=root
+spring.datasource.password=
+
+######################## MYSQL DB #######################################################
```
fviale: Maybe it's easier to use commented out configurations for various database, but why not use different files instead ?

src/main/resources/application.properties
```diff
@@ -13,17 +13,47 @@ server.contextPath=/
...
-# note the ProActiveMySQL5InnoDBDialect class that enforces the utf8mb4 charset
-#spring.datasource.driverClassName=org.mariadb.jdbc.Driver
-#spring.jpa.database-platform=org.ow2.proactive.catalog.util.ProActiveMySQL5InnoDBDialect
-#spring.datasource.url=jdbc:mariadb://localhost:3306/catalog
+spring.datasource.username=root
+spring.datasource.password=
 
-#spring.datasource.username=root
-#spring.datasource.password=
+######################## MYSQL DB ########################################################
```
fviale: same remark as above

build.gradle
nebilbenmabrouk: removed

build.gradle
nebilbenmabrouk: done and parent-bom updated

build.gradle
nebilbenmabrouk: changed to 'testCompile' and commented because Oracle jdbc dependency can be downloaded only with credentials. I just kept it there to mention the version used to test Oracle DB for future needs.

src/integration-test/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectControllerIntegrationTest.java
nebilbenmabrouk: I commented tests and forgot about them. Block comment removed and tests work fine. Sorry for that.

src/integration-test/resources/application-test.properties
nebilbenmabrouk: Using different files requires using the notion of 'profiles' in Spring. This implies to (i) hard-code the profile name just before the data source code and (ii) specify the profile at deployment time, which adds further useless constraints for the deployment of the catalog.  I prefer using a centralized properties file where we provide samples of different data sources. This is (as you already mentioned) easier for PA admin and Catalog tester.

build.gradle
```diff
@@ -204,6 +204,18 @@ dependencies {
     // library to build pdf reports
     compile 'com.github.dhorions:boxable'
 
+    //support for Oracle DB version 12.1.0.2
```
fviale: can you put a comment to explain why it is commented out, and how it can be used (credentials)?

PR #161, opened 2019-02-22
Extract depends on dependency from job and task variables

aminelouati:
In this PR, we extract depends on dependency from job and task variables and add it to the Metadata of the catalog object.

### DIFF
--- a/src/main/java/org/ow2/proactive/catalog/service/KeyValueLabelMetadataHelper.java
+++ b/src/main/java/org/ow2/proactive/catalog/service/KeyValueLabelMetadataHelper.java
@@ -29,6 +29,7 @@ import java.io.ByteArrayInputStream;
 import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
@@ -79,6 +80,16 @@ public class KeyValueLabelMetadataHelper {
     private boolean isGenericInformation(KeyValueLabelMetadataEntity keyValueLabelMetadataEntity) {
         return keyValueLabelMetadataEntity.getLabel().equals(WorkflowParser.ATTRIBUTE_GENERIC_INFORMATION_LABEL);
     }
+
+    public List<KeyValueLabelMetadataEntity>getOnlyNotDuplicatedDependsOn(List<KeyValueLabelMetadataEntity> keyValueLabelMetadataEntities) {
+        return new ArrayList<>(
+                new HashSet<>(keyValueLabelMetadataEntities.stream().filter(this::isDependsOn).collect(Collectors.toList())));
+    }
+
+    private boolean isDependsOn(KeyValueLabelMetadataEntity keyValueLabelMetadataEntity) {
+        return keyValueLabelMetadataEntity.getLabel().equals(WorkflowParser.ATTRIBUTE_DEPENDENCIES_LABEL);
+    }
+
 
     public List<KeyValueLabelMetadataEntity> extractKeyValuesFromRaw(String kind, byte[] rawObject) {
         AbstractCatalogObjectParser catalogObjectParser = parsers.stream()
--- a/src/main/java/org/ow2/proactive/catalog/service/KeyValueLabelMetadataHelper.java
+++ b/src/main/java/org/ow2/proactive/catalog/service/KeyValueLabelMetadataHelper.java
@@ -81,15 +81,16 @@ public class KeyValueLabelMetadataHelper {
         return keyValueLabelMetadataEntity.getLabel().equals(WorkflowParser.ATTRIBUTE_GENERIC_INFORMATION_LABEL);
     }
 
-    public List<KeyValueLabelMetadataEntity>getOnlyNotDuplicatedDependsOn(List<KeyValueLabelMetadataEntity> keyValueLabelMetadataEntities) {
-        return new ArrayList<>(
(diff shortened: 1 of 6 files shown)


src/test/java/org/ow2/proactive/catalog/util/ProActiveCatalogObjectParserTest.java
```diff
@@ -151,6 +167,13 @@ private String findValueForKeyAndLabel(List<KeyValueLabelMetadataEntity> result,
                      .getValue();
     }
 
+    private List<String> findValuesForKeyAndLabel(List<KeyValueLabelMetadataEntity> result, String key, String label) {
+        return result.stream()
+                     .filter(metadata -> metadata.getKey().equals(key) && metadata.getLabel().equals(label))
+                     .map(metadata -> metadata.getValue())
```
Mykhailenko:
```suggestion
                     .map(KeyValueLabelMetadataEntity::getValue)
```

src/main/java/org/ow2/proactive/catalog/util/parser/WorkflowParser.java
```diff
@@ -129,6 +142,25 @@ private void addVariableIfNotNullAndModelIfNotEmpty(
...
+        String name = taskVariable.getName();
+        String value = taskVariable.getValue();
+        String model = taskVariable.getModel();
+        if (checkIfNotNull(name, model) && checkIfNotEmpty(name, model)) {
+            addDependencyOn(keyValueMapBuilder, value, model);
+        }
+
+    }
+
+    private void addDependencyOn(ImmutableList.Builder<KeyValueLabelMetadataEntity> keyValueMapBuilder, String value,
```
ShatalovYaroslav: Filtering can be fine in KeyValueLabelMetadataHelper class.
 But do you believe that KeyValueMap can not store duplicated values? Or will not throw an error?
Maybe it's better to add check here for duplicated value before adding the duplicated Entity?

I may be wrong. Please correct me, if it works fine locally with DB on your side.
I'm afraid that with using DB it will give an error for duplicated key, label.
Because of this constraint in KeyValueLabelMetadataEntity:
`@UniqueConstraint(columnNames = { "CATALOGOBJECTREVISION",
                                                                                          "PA_KEY",
                                                                                          "LABEL" })`

src/main/java/org/ow2/proactive/catalog/util/parser/WorkflowParser.java
```diff
@@ -89,6 +97,11 @@
            .forEach((name, value) -> addGenericInformationIfNotNull(keyValueMapBuilder, name, value));
         job.getUnresolvedVariables().forEach((jobVariableName,
                 jobVariable) -> addVariableIfNotNullAndModelIfNotEmpty(keyValueMapBuilder, jobVariable));
+        for (Task task : job.getTasks()) {
+            task.getVariables().forEach((taskVariableName,
+                    taskVariable) -> addDependencyIfCatalogObjectModelExist(keyValueMapBuilder, taskVariable));
+
+        }
```
ShatalovYaroslav: Please point me where the variables are parsed on job level (not for tasks)?

src/test/resources/workflows/workflow_variables_with_catalog_object_model.xml
```diff
@@ -0,0 +1,58 @@
...
+    <variable name="called_workflow2" value="data-connectors/FTP" model="PA:Catalog_Object"/>
+    <variable name="emptyValue" value="" />
+  </variables>
+  <taskFlow>
+    <task name="Linux_Bash_Task">
+      <description>
+        <![CDATA[ The simplest task, ran by a bash engine. ]]>
+      </description>
+      <variables>
+        <variable name="called_workflow3" value="finance/QuantLib" model="PA:Catalog_Object"/>
```
ShatalovYaroslav: you can add some duplication on depends on WF from other task to test

src/main/java/org/ow2/proactive/catalog/util/parser/WorkflowParser.java
aminelouati: Line 98 and 145

src/main/java/org/ow2/proactive/catalog/service/KeyValueLabelMetadataHelper.java
```diff
@@ -80,6 +81,17 @@ private boolean isGenericInformation(KeyValueLabelMetadataEntity keyValueLabelMe
         return keyValueLabelMetadataEntity.getLabel().equals(WorkflowParser.ATTRIBUTE_GENERIC_INFORMATION_LABEL);
     }
 
+    public List<KeyValueLabelMetadataEntity>
```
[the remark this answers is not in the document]
aminelouati: yes. I added some tests for it.

src/test/java/org/ow2/proactive/catalog/util/ProActiveCatalogObjectParserTest.java
```diff
@@ -151,6 +170,15 @@ private String findValueForKeyAndLabel(List<KeyValueLabelMetadataEntity> result,
                      .getValue();
     }
 
+    private List<String> findValuesForKeyAndLabel(List<KeyValueLabelMetadataEntity> result, String key, String label) {
+        return new ArrayList<>(new HashSet<>(result.stream()
```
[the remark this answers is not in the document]
ShatalovYaroslav: or change name to findUniqueValuesForKeyAndLabel

PR #163, opened 2019-02-27
Add depends on to Metadata List

aminelouati:


src/main/java/org/ow2/proactive/catalog/util/parser/WorkflowParser.java
```diff
@@ -129,19 +144,64 @@ private void addVariableIfNotNullAndModelIfNotEmpty(
...
+                keyValueMapBuilder.add(new KeyValueLabelMetadataEntity(getNameAndBucketFromDependencyOn(value),
+                                                                       LATEST_VERSION,
+                                                                       ATTRIBUTE_DEPENDS_ON_LABEL));
+            }
+        }
+    }
+
+    private String getNameAndBucketFromDependencyOn(String calledWorkflow) {
+        try {
+            return calledWorkflow.split("/")[0] + "/" + calledWorkflow.split("/")[1];
```
fviale: please do a single split call, not a big deal, but it's prettier

src/main/java/org/ow2/proactive/catalog/util/parser/WorkflowParser.java
```diff
@@ -100,16 +101,17 @@
...
         addJobDescriptionIfNotNullAndNotEmpty(keyValueMapBuilder, job);
         addJobVizualisationIfNotNullAndNotEmpty(keyValueMapBuilder, job);
 
-        return keyValueMapBuilder.build();
+        return new ArrayList<>(keyValueMapBuilder.build());
     }
 
     private void addProjectNameIfNotNullAndNotEmpty(
-            ImmutableList.Builder<KeyValueLabelMetadataEntity> keyValueMapBuilder, Job job) {
+            ImmutableSet.Builder<KeyValueLabelMetadataEntity> keyValueMapBuilder, Job job) {
```
ShatalovYaroslav: Fabien noticed in general in case of using Set, the order is not guaranteed. I'm not sure if the order of metadata is used somewhere, but before it was ordered in List. 
We can use LinkedHashSet, but it seems that Guava ImmutableSet already provides order according to this answer 
https://stackoverflow.com/questions/25005595/class-that-describes-a-immutable-ordered-set
https://github.com/google/guava/wiki/ImmutableCollectionsExplained
and my tests. 
So we can leave like this.

src/main/java/org/ow2/proactive/catalog/util/parser/WorkflowParser.java
```diff
@@ -142,38 +144,64 @@ private void addVariableIfNotNullAndModelIfNotEmpty(
...
         }
 
     }
 
-    private void addDependencyOn(ImmutableList.Builder<KeyValueLabelMetadataEntity> keyValueMapBuilder, String value,
+    private void addDependsOn(ImmutableSet.Builder<KeyValueLabelMetadataEntity> keyValueMapBuilder, String value,
             String model) {
         if (model.equalsIgnoreCase(CATALOG_OBJECT_MODEL)) {
-            keyValueMapBuilder.add(new KeyValueLabelMetadataEntity(DEPENDS_ON, value, ATTRIBUTE_DEPENDENCIES_LABEL));
+            if (getRevisionFromDependencyOn(value).isPresent()) {
```
ShatalovYaroslav: The code readability and maintainability can be improved by using just:
     ` keyValueMapBuilder.add(new KeyValueLabelMetadataEntity(getNameAndBucketFromDependencyOn(value),
                getRevisionFromDependencyOn(value).orElse(LATEST_VERSION),
                ATTRIBUTE_DEPENDS_ON_LABEL));`

src/main/java/org/ow2/proactive/catalog/util/parser/WorkflowParser.java
ShatalovYaroslav: I agree

src/main/java/org/ow2/proactive/catalog/util/parser/WorkflowParser.java
```diff
@@ -142,38 +148,63 @@ private void addVariableIfNotNullAndModelIfNotEmpty(
...
+            if (model.equalsIgnoreCase(CATALOG_OBJECT_MODEL)) {
+                keyValueMapBuilder.add(new KeyValueLabelMetadataEntity(getNameAndBucketFromDependsOn(value),
+                        getRevisionFromDependsOn(value).orElse(LATEST_VERSION),
+                        ATTRIBUTE_DEPENDS_ON_LABEL));
+            }
+        }
+    }
+
+    private String getNameAndBucketFromDependsOn(String calledWorkflow) {
+        Pattern pattern = Pattern.compile(CATALOG_OBJECT_MODEL_REGEXP);
```
ShatalovYaroslav: Could you please extract this line `Pattern pattern = Pattern.compile(CATALOG_OBJECT_MODEL_REGEXP);`
to final private class variable?
Because it's expensive operation and it's better not to do it every time for every variable

src/main/java/org/ow2/proactive/catalog/util/parser/WorkflowParser.java
```diff
@@ -142,38 +148,63 @@ private void addVariableIfNotNullAndModelIfNotEmpty(
...
+        } else {
+            return (matcher.group(1));
+
+        }
+    }
+
+
+    private Optional<String> getRevisionFromDependsOn(String calledWorkflow) {
+        try {
+            return Optional.of(calledWorkflow.split("/")[2]);
```
ShatalovYaroslav: why don't you use your regexp pattern to extract getRevisionFromDependsOn 
with group 3 for example?

src/test/java/org/ow2/proactive/catalog/util/ProActiveCatalogObjectParserTest.java
```diff
@@ -86,19 +84,20 @@ public void testParseWorkflowWithModelVariables() throws Exception {
...
-        assertThat(findUniqueValuesForKeyAndLabel(result,
-                                                  "depends_on",
-                                                  "dependencies")).contains("deep-learning-workflows/Custom_Sentiment_Analysis_In_Bing_News");
-        assertThat(findUniqueValuesForKeyAndLabel(result, "depends_on", "dependencies")).hasSize(4);
+        assertThat(dependsOnKeys).contains("basic-examples/Native_Task");
+        assertThat(dependsOnKeys).contains("data-connectors/FTP");
+        assertThat(dependsOnKeys).contains("finance/QuantLib");
+        assertThat(dependsOnKeys).contains("deep-learning-workflows/Custom_Sentiment_Analysis_In_Bing_News");
+        assertThat(dependsOnKeys).contains("data-connectors/SFTP");
+        assertThat(dependsOnKeys).contains("data-connectors/URL");
```
ShatalovYaroslav: could you add one assert for revision number not LATEST?

ShatalovYaroslav:
Nice, please squash and merge

PR #170, opened 2019-04-01
Call graph

aminelouati:
In this PR we,

1. Added a new endpoint called `getCatalogObjectsOfKindWorkflow` that lists all catalog objects having *workflow* as a kind and the corresponding integration test.
2. Implemented the business logic of the dependency graph. 
3. Added a new endpoint called `getCallGraph` that returns a jpg image containing the dependency graph of the entire Catalog and many integration tests.


src/main/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectController.java
```diff
@@ -160,6 +160,13 @@ public CatalogObjectMetadataList create(
         return catalogObjectService.getContentTypes();
     }
 
+    @ApiOperation(value = "Lists all catalog objects of kind workflow")
+    @RequestMapping(value = "/workflows", method = GET, produces = "application/json")
+    @ResponseStatus(HttpStatus.OK)
+    public List<Object[]> listWorkflows() {
```
ShatalovYaroslav: Sorry, I don't agree with this API End-point. First you already have API to get buckets by kind:
/bucket-controller/listUsingGET
And also you have agreed way to get objects by kind from specified bucket:
/catalog-object-controller/listUsingGET_1

Maybe this approach is not good for your use-case and you really need to get all workflows from all buckets in 1 call, then please generalize your endpoint. We need to keep catalog generic and not only for specific cases like workflow kind.

1) please put in response List of CatalogObjectMetadata (now the list of "cos.bucket.bucketName, cos.id.name" )
2) put kind as parameter of method

This requires refactoring the underlying calls from this method.

fviale: I agree with @ShatalovYaroslav that the catalog should be generic, and so the endpoint should be generic. Besides that, we plan also to give the possibility to find the call graph of scripts.

aminelouati: Thank you @ShatalovYaroslav and @fviale for the reviews but I think that it is a mistake to include this endpoint in my PR. It is not related to the call graph work. Actually, it is a request made by @amouhoub for job analysis. The main need for this is to create an auto-complete feature in the job analytics to allow to type workflow names faster. So we'll only need workflow names and bucket names.
If we include all the metadata, the response will way bigger for what we need... No other portal has needed a listing of all CatalogObjectMetadata for the moment and clearly there is no need for that or it would have been done already. 
I will remove this part from this PR and create another one which consider your reviews.

aminelouati:
re-test this please

src/main/java/org/ow2/proactive/catalog/callgraph/CallGraphHolder.java
```diff
@@ -0,0 +1,97 @@
...
+
+    }
+
+    public boolean removeNode(String bucketName, String objectName) {
+        return dependencyGraph.removeVertex(new GraphNode.Builder().bucketName(bucketName)
+                                                                   .objectName(objectName)
+                                                                   .build());
+    }
+
+    public DefaultEdge addDependsOnDependency(GraphNode graphNode1, GraphNode graphNode2) {
```
nebilbenmabrouk: Can be renamed addDependsOnEdge

aminelouati: you are right. I will change the name.

mboussaa:
Left some live comments. LGTM

src/main/java/org/ow2/proactive/catalog/callgraph/GraphNode.java
```diff
@@ -0,0 +1,101 @@
...
+
+import org.objectweb.proactive.annotation.PublicAPI;
+
+
+/**
+ * @author ActiveEon Team
+ * @since 2019-03-22
+ */
+
+@PublicAPI
```
ShatalovYaroslav: you can use lombok.Builder; instead of writing your own

PR #176, opened 2019-04-18
fix call graph bug using JGraphx in Headless Environment

aminelouati:


src/main/java/org/ow2/proactive/catalog/callgraph/CatalogObjectCallGraphPDFGenerator.java
```diff
@@ -79,8 +81,6 @@
 
     private static final String MAIN_TITLE = "ProActive Call Graph Report";
 
-    private CallGraphHolder callGraphHolder = new CallGraphHolder();
```
mcompastie: You should update in accordance the tests of this class so you can include CallGraphHolder class removal in part of this pull request.

src/main/java/org/ow2/proactive/catalog/callgraph/CatalogObjectCallGraphPDFGenerator.java
```diff
@@ -263,4 +266,25 @@ private void buildCatalogCallGraph(List<CatalogObjectMetadata> catalogObjectMeta
 
     }
 
+    private class mxGraphComponentWithoutDragAndDrop extends mxGraphComponent {
```
mboussaa: Java doc would be very helpful to understand the new added class

src/main/java/org/ow2/proactive/catalog/callgraph/CatalogObjectCallGraphPDFGenerator.java
mcompastie: Ok, I hadn't seen the new occurrence to that class introduced by your commit. I validate your PR.

src/main/java/org/ow2/proactive/catalog/callgraph/CatalogObjectCallGraphPDFGenerator.java
MaelAudren: I agree with Mohamed it would be good to add comments explaining why those two class were added.

src/main/java/org/ow2/proactive/catalog/callgraph/CatalogObjectCallGraphPDFGenerator.java
```diff
@@ -263,4 +266,29 @@ private void buildCatalogCallGraph(List<CatalogObjectMetadata> catalogObjectMeta
...
+            return new mxGraphHandlerWithoutDragAndDrop(this);
+        }
+    }
+
+    private class mxGraphHandlerWithoutDragAndDrop extends mxGraphHandler {
+        public mxGraphHandlerWithoutDragAndDrop(mxGraphComponent graphComponent) {
+            super(graphComponent);
+        }
+
+        protected void installDragGestureHandler() {
```
MaelAudren: @aminelouati we forgot to put the override annotation here

PR #177, opened 2019-04-18
Add a new endpoint getCallGraphForSelectedObjects

aminelouati:
In this PR we:
1. Added a new endpoint getCallGraphForSelectedObjects to et the call graph of selected catalog objects in a bucket. 
2. Some integration tests

### DIFF
--- a/src/main/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectCallGraphController.java
+++ b/src/main/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectCallGraphController.java
@@ -120,23 +120,10 @@ public class CatalogObjectCallGraphController {
         restApiAccessService.checkAccessBySessionIdForBucketAndThrowIfDeclined(sessionIdRequired,
                                                                                sessionId,
                                                                                bucketName);
-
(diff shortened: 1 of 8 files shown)


src/main/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectCallGraphController.java
```diff
@@ -92,14 +94,52 @@ public void getCallGraph(HttpServletResponse response,
...
+            flushResponse(response, content);
+
+        } else {
+
+            byte[] content = catalogObjectCallGraphService.generateBytesCallGraph(Collections.singletonList(bucketName),
+                                                                                  kind,
+                                                                                  contentType);
+            flushResponse(response, content);
+
+        }
```
MaelAudren: Replace if...else... condition by:
```java
            byte[] content = catalogObjectNames
.map(names->catalogObjectCallGraphService.generateBytesCallGraph(bucketName,names,kind,contentType)
.orElse(catalogObjectCallGraphService.generateBytesCallGraph(Collections.singletonList(bucketName), kind,contentType));
            flushResponse(response, content);
```

PR #179, opened 2019-04-25
consider script dependencies in the call graph

aminelouati:


src/integration-test/java/org/ow2/proactive/catalog/service/CatalogObjectServiceIntegrationTest.java
```diff
@@ -476,6 +476,74 @@ public void testCatalogObjectNotFound() throws IOException {
...
+        assertThat(dependsOnKeys).contains("cloud-automation-scripts/Service_Start");
+        assertThat(dependsOnKeys).contains("cloud-automation-scripts/Pre_Trigger_Action");
+        assertThat(dependsOnKeys).contains("scripts/check_ip");
+        assertThat(dependsOnKeys).contains("notification-tools/Web_Validation_Script");
+        assertThat(dependsOnKeys).contains("scripts/is_mac");
+        assertThat(dependsOnKeys).contains("scripts/update_variables_from_system");
+        assertThat(dependsOnKeys).contains("scripts/update_variables_from_file");
+        assertThat(dependsOnKeys).hasSize(8);
+
+        //'bucket/A_Object' depends on bucket/B_Object'
```
mbenguig: SImply for me, what does it mean "xx depends on yy" ?

src/main/java/org/ow2/proactive/catalog/util/parser/WorkflowParser.java
```diff
@@ -84,12 +87,18 @@
 
     private static final String CATALOG_OBJECT_MODEL_REGEXP = "^([^/]+/[^/]+)(/[^/][0-9]{12})?$";
 
+    private static final String SCRIPT_URL_REGEX = "^((http|https)://.*|\\$\\{PA\\_CATALOG\\_REST\\_URL\\})/buckets/([^/]+)/resources/([^/]+)(/revisions/([^/][0-9]{12}))?/raw";
```
mbenguig: If you have an example of SCRIPT_URL_REGEX , could you put it above as a comment? (the same for the other regexp)

src/integration-test/java/org/ow2/proactive/catalog/service/CatalogObjectServiceIntegrationTest.java
```diff
@@ -476,6 +476,74 @@ public void testCatalogObjectNotFound() throws IOException {
...
+        //check the correctness of the DB
+        long firstCommitTimeDependsOn = catalogObjectMetadata.getCommitDateTime()
+                                                             .atZone(ZoneId.systemDefault())
+                                                             .toInstant()
+                                                             .toEpochMilli();
+        CatalogObjectDependencies catalogObjectDependencyListOfAObjectFirstCommit = catalogObjectService.getObjectDependencies(bucketName,
+                                                                                                                               workflowName,
+                                                                                                                               firstCommitTimeDependsOn);
+        List<String> BucketAndObjectNameDependsOnListOfAObjectFromDBFirstCommit = catalogObjectDependencyListOfAObjectFirstCommit.getDependsOnList()
```
mbenguig: BucketAndObjectNameDependsOnListOfAObjectFromDBFirstCommit  -> bucketAndObjectNameDependsOnListOfAObjectFromDBFirstCommit

src/integration-test/java/org/ow2/proactive/catalog/service/CatalogObjectServiceIntegrationTest.java
aminelouati: It means that xx is calling yy. It can be throw PA:CATALOG_OBJECT model (submitFromCatalog) or a script url.

src/main/java/org/ow2/proactive/catalog/util/parser/WorkflowParser.java
aminelouati: Ok

fviale: scripts can be defined by other protocols than http(s), e.g. file:// ftp://

src/main/java/org/ow2/proactive/catalog/util/parser/WorkflowParser.java
```diff
@@ -167,6 +177,79 @@ private void addDependsOnIfCatalogObjectModelExistOnTaskVariable(
...
+
+        keyValueMapBuilder.add(new KeyValueLabelMetadataEntity(getNameAndBucketFromUrl(scriptUrl),
+                                                               getRevisionFromUrl(scriptUrl).orElse(LATEST_VERSION),
+                                                               ATTRIBUTE_DEPENDS_ON_LABEL));
+
+    }
+
+    private String getNameAndBucketFromUrl(String scriptUrl) {
+        Matcher matcher = URL_PATTERN.matcher(scriptUrl);
+        if (!(scriptUrl.matches(SCRIPT_URL_REGEX) && matcher.find())) {
```
fviale: I don't think you should throw an exception here, but rather, only add depends on when the pattern matches

src/main/java/org/ow2/proactive/catalog/util/parser/WorkflowParser.java
fviale: I think the pattern should only try to detect catalog urls

src/main/java/org/ow2/proactive/catalog/util/parser/WorkflowParser.java
aminelouati: Ok

src/main/java/org/ow2/proactive/catalog/util/parser/WorkflowParser.java
aminelouati: In that case, can we accept such URL ? as it is done in the tests ? 
http://localhost:8080/catalog/buckets/cloud-automation-scripts/Service_Start/raw

aminelouati: if I understood well you suggest to detect only catalog urls (${PA_CATALOG_REST_URL}) and remove http|https|file|ftp ?. If it is the case, selection script url will not pass in the tests. Should I remove these tests then ?
Thanks

fviale: I don't quite understand, maybe you can explain me on monday.

Indeed, I suggest that we only try to parse catalog urls.

PR #181, opened 2019-05-06
adapt call graph visualization for large graph

aminelouati:


src/main/java/org/ow2/proactive/catalog/callgraph/CatalogObjectCallGraphPDFGenerator.java
```diff
@@ -162,6 +177,34 @@
         }
     }
 
+    private void drawPartitionImageOnSinglePage(BufferedImage partition, PDDocument doc, int pageIndex)
+            throws IOException {
+        PDPage page = doc.getPage(0);
+        if (pageIndex != 0) {
+            page = addNewPage(doc);
+        }
```
mboussaa: (suggestion) PDPage page = pageIndex != 0 ? addNewPage(doc) : doc.getPage(0);

PR #186, opened 2019-06-12
Add content-type filter to listCatalogObjectNameReference endpoint

aminelouati:


src/main/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectController.java
```diff
@@ -162,15 +161,19 @@ public CatalogObjectMetadataList create(
         return catalogObjectService.getContentTypes();
     }
 
-    @ApiOperation(value = "Lists catalog object name references by kind")
+    @ApiOperation(value = "Lists catalog object name references by kind and content type")
```
amouhoub: content-type

src/main/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectController.java
```diff
@@ -162,15 +161,19 @@ public CatalogObjectMetadataList create(
...
     @ApiResponses(value = { @ApiResponse(code = 401, message = "User not authenticated"),
                             @ApiResponse(code = 403, message = "Permission denied") })
     @RequestMapping(value = "/references", method = GET, produces = "application/json")
     @ResponseStatus(HttpStatus.OK)
     public List<CatalogObjectNameReference> listCatalogObjectNameReference(
             @ApiParam(value = "sessionID", required = false) @RequestHeader(value = "sessionID", required = false) String sessionId,
-            @ApiParam(value = "kind") @RequestHeader(value = "kind", required = false) Optional<String> kind) {
-        return catalogObjectService.getAccessibleCatalogObjectsNameReferenceByKind(sessionIdRequired, sessionId, kind);
+            @ApiParam(value = "Filter according to kind") @RequestHeader(value = "kind", required = false) Optional<String> kind,
+            @ApiParam(value = "Filter according to content type") @RequestHeader(value = "contentType", required = false) Optional<String> contentType) {
```
amouhoub: content-type

src/main/java/org/ow2/proactive/catalog/service/CatalogObjectService.java
```diff
@@ -586,23 +586,29 @@ public CatalogObjectMetadata restore(String bucketName, String name, Long commit
...
 
-    public List<CatalogObjectNameReference> getAccessibleCatalogObjectsNameReferenceByKind(boolean sessionIdRequired,
-            String sessionId, Optional<String> kind) {
-        List<CatalogObjectNameReference> catalogObjectsNameReferenceByKind;
-        if (!kind.isPresent()) {
-            catalogObjectsNameReferenceByKind = generateCatalogObjectsNameReferenceByKind(catalogObjectRepository.findAllCatalogObjectNameReferenceByKind());
+    public List<CatalogObjectNameReference> getAccessibleCatalogObjectsNameReferenceByKindAndContentType(
+            boolean sessionIdRequired, String sessionId, Optional<String> kind, Optional<String> contentType) {
+        List<CatalogObjectNameReference> catalogObjectsNameReferenceByKindAndContentType;
+        if (kind.isPresent() && contentType.isPresent()) {
```
ShatalovYaroslav: All the following part can be improved/replaced by just calling one method:
 `
 catalogObjectsNameReferenceByKindAndContentType = generateCatalogObjectsNameReferenceByKind(
                catalogObjectRepository.findCatalogObjectNameReferenceByKindAndContentType(
                        kind.orElse(""),
                        contentType.orElse("")));
`
I've checked that in case of Optional empty it will pass "" String as argument. It works fine for `LIKE lower(concat('%', ?2, '%'))"`
In general try to avoid many if-else conditions. It makes the design more clear and easier to maintain.

src/main/java/org/ow2/proactive/catalog/repository/CatalogObjectRepository.java
```diff
@@ -52,9 +52,15 @@
...
     Set<String> findAllContentTypes();
 
+    @Query(value = "SELECT cos.bucket.bucketName, cos.id.name FROM CatalogObjectEntity cos WHERE lower(cos.kind) LIKE lower(concat('%', ?1, '%')) AND lower(cos.contentType) LIKE lower(concat('%', ?2, '%'))")
+    List<Object[]> findCatalogObjectNameReferenceByKindAndContentType(String kind, String contentType);
+
     @Query(value = "SELECT cos.bucket.bucketName, cos.id.name FROM CatalogObjectEntity cos WHERE lower(cos.kind) LIKE lower(concat('%', ?1, '%'))")
     List<Object[]> findCatalogObjectNameReferenceByKind(String kind);
 
+    @Query(value = "SELECT cos.bucket.bucketName, cos.id.name FROM CatalogObjectEntity cos WHERE lower(cos.contentType) LIKE lower(concat('%', ?1, '%'))")
+    List<Object[]> findCatalogObjectNameReferenceByContentType(String contentType);
```
ShatalovYaroslav: this method can be removed if you will use just general version with kind = ""
findCatalogObjectNameReferenceByKindAndContentType

src/main/java/org/ow2/proactive/catalog/repository/CatalogObjectRepository.java
```diff
@@ -52,9 +52,15 @@
     @Query(value = "SELECT DISTINCT cos.contentType FROM CatalogObjectEntity cos")
     Set<String> findAllContentTypes();
 
+    @Query(value = "SELECT cos.bucket.bucketName, cos.id.name FROM CatalogObjectEntity cos WHERE lower(cos.kind) LIKE lower(concat('%', ?1, '%')) AND lower(cos.contentType) LIKE lower(concat('%', ?2, '%'))")
+    List<Object[]> findCatalogObjectNameReferenceByKindAndContentType(String kind, String contentType);
+
     @Query(value = "SELECT cos.bucket.bucketName, cos.id.name FROM CatalogObjectEntity cos WHERE lower(cos.kind) LIKE lower(concat('%', ?1, '%'))")
     List<Object[]> findCatalogObjectNameReferenceByKind(String kind);
```
ShatalovYaroslav: this method can be removed if you will use just general version with contentType = ""
findCatalogObjectNameReferenceByKindAndContentType

src/main/java/org/ow2/proactive/catalog/repository/CatalogObjectRepository.java
amouhoub: @ShatalovYaroslav I don't know how spring works but in SQL if you have a where clause with `kind=""` then you require a join that satisfies this condition, which will always return 0 results as all catalog objects have a kind.

amouhoub: @aminelouati says it works so I guess Spring does some auto-mapping and transforms kind="" to kind="*"

amouhoub: Nice catch @ShatalovYaroslav  ;)

PR #196, opened 2019-07-23
Improve catalog

aminelouati:
In this PR, we added a new attribute contentCount to the BucketMetadata class to improve the performance.

### DIFF
--- a/src/main/java/org/ow2/proactive/catalog/service/BucketService.java
+++ b/src/main/java/org/ow2/proactive/catalog/service/BucketService.java
@@ -140,11 +140,7 @@ public class BucketService {
                                                                                                                  objectName.orElse("")));
 
         } else {
-            entities = bucketRepository.findAll()
-                                       .stream()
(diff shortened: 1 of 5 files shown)


src/main/java/org/ow2/proactive/catalog/service/BucketService.java
```diff
@@ -94,47 +94,61 @@ public BucketMetadata getBucketMetadata(String bucketName) {
...
-            entities = bucketRepository.findContainingKindAndContentType(kind.orElse(""),
-                                                                         contentType.orElse(""),
-                                                                         objectName.orElse(""));
+            entities = generateBucketMetadataList(bucketRepository.findContainingKindAndContentTypeAndObjectName(kind.orElse(""),
+                                                                                                                 contentType.orElse(""),
+                                                                                                                 objectName.orElse("")));
+
         } else {
-            entities = bucketRepository.findAll();
+            entities = bucketRepository.findAll()
```
ShatalovYaroslav: Could you use generateBucketMetadataList() here as well?

PR #199, opened 2019-08-28
use microservices-common to manage the exceptions

MaelAudren:


build.gradle
```diff
@@ -87,7 +87,7 @@ licenseReport {
 
 repositories {
 
-    if (project.hasProperty('local')) {
+   if (project.hasProperty('local')) {
```
tao-xinxiu: it seems missing a space here.

src/integration-test/resources/archives/workflow_0.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
tao-xinxiu: The file could be renamed more meaningful as "workflow_hello_world" ?

build.gradle
MaelAudren: Done

src/integration-test/resources/archives/workflow_0.xml
MaelAudren: I updated the code to remove this file.

src/main/java/org/ow2/proactive/catalog/rest/exceptiohandler/ExceptionHandlerController.java
```diff
@@ -49,33 +43,16 @@
 @Controller
 @ControllerAdvice
 @Log4j
-public class ExceptionHandlerController {
-
-    @ExceptionHandler(Exception.class)
```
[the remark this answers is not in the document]
MaelAudren: ExceptionHandlerController extends ExceptionHandlerAdvice which have a general exception handler.

PR #204, opened 2020-01-31
Add feature to update bucket group or owner 

ShatalovYaroslav:


src/main/java/org/ow2/proactive/catalog/service/BucketService.java
```diff
@@ -83,6 +90,25 @@ public BucketMetadata createBucket(String name, String owner) throws DataIntegri
...
+        createRevisionForObjects(bucketName, COMMIT_MESSAGE_UPDATE_BUCKET);
+
+        return new BucketMetadata(bucketEntity, bucketEntity.getCatalogObjects().size());
+    }
+
+    //create a new revision for objects when the bucket owner is updated
+    protected void createRevisionForObjects(String bucketName, String commitMessage) {
+        List<CatalogObjectRevisionEntity> objectsList = catalogObjectService.listCatalogObjectsEntities(Arrays.asList(bucketName));
+
+        objectsList.forEach(obj -> catalogObjectService.createCatalogObjectRevision(obj, commitMessage));
```
aminelouati: Does this new revision will appear in the revisions list panel in the Catalog Portal ? Otherwise, what is the purpose to create an object revision ?
Thanks

src/test/java/org/ow2/proactive/catalog/service/BucketServiceTest.java
```diff
@@ -100,6 +104,34 @@ public void testCreateBucket() throws Exception {
...
+        CatalogObjectRevisionEntity catalogObjectRevisionEntity = CatalogObjectRevisionEntity.builder()
+                                                                                             .catalogObject(CatalogObjectEntity.builder()
+                                                                                                                               .kind(SupportedParserKinds.WORKFLOW.toString())
+                                                                                                                               .build())
+                                                                                             .build();
+        objectsList.add(catalogObjectRevisionEntity);
+
+        when(bucketRepository.findOneByBucketName(anyString())).thenReturn(mockedBucket);
+        when(bucketRepository.save(mockedBucket)).thenReturn(mockedBucketWithOwner);
+        when(catalogObjectService.listCatalogObjectsEntities((Arrays.asList("BUCKET-NAME-TEST")))).thenReturn(objectsList);
```
aminelouati: You can set "BUCKET-NAME-TEST" as a constant

src/main/java/org/ow2/proactive/catalog/service/BucketService.java
ShatalovYaroslav: yes, it will appear as the normal new revision. This is the way how to update the key-value metadata values, as they are connected with revisions.

PR #222, opened 2020-09-09
Add contentType filter to graphql API

aminelouati:


ShatalovYaroslav (approved):

ShatalovYaroslav:
Could you add real integration test with the graphql query  to be sure, that functionality works as expected?
You should find some examples of such tests with queries.

AlexFalessi (approved):

PR #230, opened 2021-02-25
synchronize the project name between the frontend and the backend (workflow xml)

aminelouati:


amouhoub (approved):

src/main/java/org/ow2/proactive/catalog/service/WorkflowXmlManipulator.java
```diff
@@ -111,6 +133,10 @@ private Element replaceJobName(Element element, String jobName) {
         return replaceAttributeValue(element, "name", jobName);
     }
 
+    private Element replaceOrAddProjectName(Element rootElement, String projectName) {
```
nebilbenmabrouk, line 136: Check that projectName is an empty space :)

src/main/java/org/ow2/proactive/catalog/service/WorkflowXmlManipulator.java
```diff
@@ -82,6 +82,28 @@
         }
     }
 
+    public byte[] replaceOrAddProjectNameOnJobLevel(final byte[] xmlWorkflow, String projectName) {
```
nebilbenmabrouk, line 85: Check that projectName is an empty space, if it is empty, do nothing

aminelouati: The projectName value (empty string or not) provided by the user will be synchronized in the underlying xml file.

src/main/java/org/ow2/proactive/catalog/service/WorkflowXmlManipulator.java
aminelouati: projectName can be empty string

src/main/java/org/ow2/proactive/catalog/service/WorkflowXmlManipulator.java
fviale: if the parameter projectName is empty, don't we expect to actually remove the attribute (this is the behavior in the studio) ?

aminelouati: Okey. I will update the PR

fviale (approved):
LGTM, to be tested live :-)

PR #242, opened 2021-09-27
list catalog objects metadata for a given bucket using multiple kind

aminelouati:


src/main/java/org/ow2/proactive/catalog/repository/CatalogObjectRevisionRepositoryImpl.java
```diff
@@ -0,0 +1,101 @@
...
+public class CatalogObjectRevisionRepositoryImpl implements CatalogObjectRevisionCustom {
+
+    @PersistenceContext
+    EntityManager em;
+
+    @Override
+    public List<CatalogObjectRevisionEntity> findDefaultCatalogObjectsOfKindListAndContentTypeAndObjectNameInBucket(
+            List<String> bucketNames, List<String> kindList, String contentType, String objectName, int pageNo,
+            int pageSize) {
+        return em.createQuery(buildCriteriaQuery(null, kindList, contentType, objectName))
```
AlexFalessi, line 48: Don't you want to pass `bucketNames` intead of `null`

AlexFalessi (approved):

mboussaa (approved):

src/main/java/org/ow2/proactive/catalog/repository/CatalogObjectRevisionRepositoryImpl.java
marounkoussaifi: I agree with Alex since you have it in the function attributes why not passing it.

aminelouati: Thanks for the review. You were right

PR #243, opened 2021-10-20
Add variables order map to the Json response

marounkoussaifi:


src/main/java/org/ow2/proactive/catalog/dto/CatalogObjectMetadata.java
```diff
@@ -143,4 +151,114 @@ private String getProjectNameIfExistsOrEmptyString() {
...
+        // This item corresponds to the variables that have no groups - They must be added first
+        groups.add("");
+
+        // Get all variables with groups
+        groups.addAll(metadataList.stream()
+                                  .filter(metadata -> metadata.getLabel().equals("variable_group") &&
+                                                      !metadata.getValue().equals("") && metadata.getValue() != null)
+                                  .map(Metadata::getValue)
+                                  .collect(Collectors.toCollection(LinkedHashSet::new)));
+
```
fviale, line 168: you should use the following map to improve the search:
`Map<String, String> variablesToGroups = metadataList.stream().filter(metadata -> metadata.getLabel().equals("variable_group") &&
                                                      !metadata.getValue().equals("") && metadata.getValue() != null).collect(Collectors.toMap(metadata -> metadata.getKey(), metadata -> metadata.getValue()));`

using this, checkIfTheVariableHasNoGroup becomes `!variablesToGroups.containsKey(variableName)`

fviale: you can also use this map to find all variables belonging to a group

PR #245, opened 2021-12-01
Add grant status for buckets and objects in endpoints' response

marounkoussaifi:


src/main/java/org/ow2/proactive/catalog/dto/BucketMetadata.java
```diff
@@ -82,6 +85,10 @@ public String getName() {
         return name;
     }
 
+    public void setRights(String bucketGrantAccessType) {
+        this.rights = bucketGrantAccessType;
```
fviale, line 89: could you change the parameter name also ?

fviale (approved):
LGTM

fviale:
Retest this please

src/main/java/org/ow2/proactive/catalog/service/BucketGrantService.java
```diff
@@ -407,4 +403,31 @@ public void deleteAllBucketGrants(long bucketId) {
...
+        String accessType = "";
+        if (user.getName().equals(bucketOwner) || user.getGroups().contains(bucketOwner.substring(6)) ||
+            bucketOwner.equals("GROUP:public-objects")) {
+            accessType = admin.toString();
+        } else {
+            long bucketId = bucketRepository.findOneByBucketName(bucketName).getId();
+            List<BucketGrantMetadata> grants = this.getAllAssignedGrantsForUserAndHisGroups(user.getName(),
+                                                                                            user.getGroups());
+            grants = grants.stream().filter(grant -> grant.getBucketId() == bucketId).collect(Collectors.toList());
+            if (grants.size() > 0) {
```
aminelouati, line 418: is it possible that grants is Null ? If yes, can you add a check here ?

marounkoussaifi: The way I did getAllAssignedGrants..., it will return an empty list in case of null or no grants.

fviale:
Retest this please

PR #246, opened 2021-12-01
Improve performance of search by kind or contentType

fviale:
- added columns kindLower and contentTypeLower
 - also added nameLower. It will not improve performances because we use like %name% expression. Some databases may have better implementation of such expressions.
 - At database startup, populate missing lowercase columns


src/main/java/org/ow2/proactive/catalog/service/DatabaseBackwardCompatibilityManager.java
```diff
@@ -40,10 +40,12 @@
     @Autowired
     ProjectNameStartupAdder projectNameStartupAdder;
 
+    @Autowired
+    NameKindContentTypeLowerCaseStartupAdder nameKindContentTypeLowerCaseStartupAdder;
+
     @PostConstruct
     public void initSetProjectNameInCatalogObjectRevisionEntity() {
```
aminelouati, line 47: maybe rename this method to cover both synchronization

aminelouati (approved):

src/main/java/org/ow2/proactive/catalog/service/DatabaseBackwardCompatibilityManager.java
fviale: ok

tao-xinxiu (approved):

PR #249, opened 2022-01-04
Fix bugs in positive grants features

marounkoussaifi:


src/main/java/org/ow2/proactive/catalog/rest/controller/BucketController.java
```diff
@@ -212,7 +212,14 @@ public BucketMetadata getMetadata(
                 String bucketGrantAccessType = bucketGrantService.getHighestGrantAccessTypeFromBucketGrants(user,
                                                                                                             data.getName(),
                                                                                                             data.getOwner());
-                BucketMetadata metadata = new BucketMetadata(data.getName(), data.getOwner(), data.getObjectCount());
+                int objectCount;
```
marounkoussaifi, line 215: This addition is to show exactly how many object are accessible for the user inside a bucket in case of object grants only or show the number of objects in case of bucket grant only (or bucket and object grant combined since this combination will allow the user to access all objects).
Before the number of all objects was showing all the tile regarding the case of only object grants.

fviale: it's kind of weird that if grantCount is 0 you return the total number of objects.

For example, a user could have access to a bucket containing one object, but no access to this object. In which case, it should display 0 and not the total object count

marounkoussaifi: When you reach this verification, it's mean you already have a grant access to the bucket or a default access if you are the owner. I can  improve it to return directly the correct number. However the no access will be added when the negative grants are implemented.

src/main/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectController.java
```diff
@@ -426,7 +426,19 @@ public CatalogObjectDependencies getDependencies(
                                                                                                pageSize);
 
             String accessType = "";
-            if (!grants.isEmpty()) {
+            boolean isBucketGrantExist = true;
```
fviale, line 429: the enclosing method is too big. Could you refactor the code a bit and add private submethods with understandable names ?

src/main/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectController.java
```diff
@@ -490,10 +502,22 @@ public CatalogObjectMetadata delete(
             // Check Grants
             AuthenticatedUser user = restApiAccessService.getUserFromSessionId(sessionId);
             if (!restApiAccessService.isBucketAccessibleByUser(sessionIdRequired, sessionId, bucketName)) {
-                if (!bucketGrantService.isTheUserGrantSufficientForTheCurrentTask(user, bucketName, admin.toString()) &&
-                    !catalogObjectGrantService.isTheUserGrantSufficientForTheCurrentTask(user,
+                List<CatalogObjectGrantMetadata> grants = catalogObjectGrantService.getAllAssignedCatalogObjectGrantsForUser(user)
```
fviale, line 505: this code should be refactored into a private method with understandable name

src/main/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectGrantController.java
```diff
@@ -91,14 +99,26 @@ public CatalogObjectGrantMetadata createCatalogObjectGrantForAUser(
             }
             // Check Grants
             if (!restApiAccessService.isBucketAccessibleByUser(sessionIdRequired, sessionId, bucketName)) {
-                if (!catalogObjectGrantService.checkInCatalogObjectGrantsIfTheUserOrUserGroupHasAdminRightsOverTheCatalogObject(user,
-                                                                                                                                bucketName,
-                                                                                                                                catalogObjectName)) {
-                    throw new CatalogObjectGrantAccessException(bucketName, catalogObjectName);
+                List<CatalogObjectGrantMetadata> grants = catalogObjectGrantService.getAllAssignedCatalogObjectGrantsForUser(user)
```
fviale, line 102: same remark as above these stream filter collect are difficult to read. Refactor using a private method with understandable name

src/main/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectGrantController.java
```diff
@@ -134,14 +154,26 @@ public CatalogObjectGrantMetadata createCatalogObjectGrantForAGroup(
...
```
fviale, line 157: same remark as above

src/main/java/org/ow2/proactive/catalog/service/BucketGrantService.java
```diff
@@ -444,6 +456,56 @@ public String getHighestGrantAccessTypeFromBucketGrants(AuthenticatedUser user,
...
 
+    /**
+     *
+     * Count the objects that are accessible for the user via the bucket or object grants
+     *
+     * @param user authenticated user
+     * @param bucketName bucket name
+     * @return the number of user's accessible objects inside the bucket
+     */
+    public int getNumberOfAccessibleObjectsInTheBucket(AuthenticatedUser user, String bucketName) {
```
fviale, line 467: same remarks as above, refactor using understandable private methods

src/main/java/org/ow2/proactive/catalog/service/CatalogObjectGrantService.java
```diff
@@ -496,15 +510,24 @@ public boolean isTheUserGrantSufficientForTheCurrentTask(AuthenticatedUser user,
...
-        grantEntities.removeIf(catalogObjectGrantEntity -> ((!catalogObjectGrantEntity.getGrantee()
-                                                                                      .equals(user.getName()) &&
-                                                             catalogObjectGrantEntity.getGrantee().equals("user")) ||
-                                                            (!user.getGroups()
-                                                                  .contains(catalogObjectGrantEntity.getGrantee()) &&
-                                                             catalogObjectGrantEntity.getGrantee().equals("group"))));
-
-        grantEntities.removeIf(catalogObjectGrantEntity -> catalogObjectGrantEntity.getCatalogObjectRevisionEntity()
-                                                                                   .getId() != catalogObjectId);
+        List<CatalogObjectGrantEntity> grantEntitiesToRemove = new LinkedList<>();
```
fviale, line 513: why is this variable called grantEntitiesToRemove ?

src/main/java/org/ow2/proactive/catalog/service/exception/CatalogObjectGrantAccessException.java
```diff
@@ -34,7 +34,7 @@
 public class CatalogObjectGrantAccessException extends ClientException {
 
     public CatalogObjectGrantAccessException(String bucketName, String catalogObjectName) {
-        super("You don't admin grant rights over the catalog object: " + catalogObjectName + " in the bucket: " +
+        super("You do not have admin grant rights over the catalog object: " + catalogObjectName + " in the bucket: " +
```
fviale, line 37: You do not have admin rights over the catalog object

src/main/java/org/ow2/proactive/catalog/rest/controller/BucketController.java
fviale: I understand, but to avoid forgetting it when you introduce the noaccess right, I would suggest to return the correct value.

At any rate, you save an extra call to getObjectCount(), which seems unnecessary. Unless this call is required in some cases ?

src/main/java/org/ow2/proactive/catalog/service/CatalogObjectGrantService.java
marounkoussaifi: Before this section there is a function called findCatalogObjectGrantEntitiesByBucketEntityId that return all object grants inside a bucket. In grantEntitesToRemove, I put all the grants that do not belong to the current user for the current object. Then I remove them from the grant list. list.removeAll(grantEntitiesToRemove).
I tried to use a stream and a filter but i was not working properly.

fviale: ok, then maybe you can refactor this code to a private method like filterGrantsThatDoesNotApplyToUser

src/main/java/org/ow2/proactive/catalog/rest/controller/BucketController.java
marounkoussaifi: Ok I'm fixing it. Usually grant count is 0 when there is no objects grants assigned to the bucket or if there is a bucket grant for the bucket. So when grant count is not 0, it means the user has access to the current bucket from object grants only, therefore, I return only the number of accessible objects. If if it's not the cas I return the number of all objects because the bucket is either public or accessible via bucket grant.

I'll try to fix it with out using if else. Directly return the correct number.

fviale: thanks

src/main/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectGrantController.java
```diff
@@ -91,14 +99,26 @@ public CatalogObjectGrantMetadata createCatalogObjectGrantForAUser(
             }
```
AlexFalessi, line 99: Looks like we can't comment on code that we expended in the PR.

I know it was added before but please change thrown exception `DataIntegrityViolationException` for `PublicBucketGrantAccessException` everywhere we try to add grant to a public bucket.

src/main/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectController.java
```diff
@@ -426,7 +426,19 @@ public CatalogObjectDependencies getDependencies(
                                                                                                pageSize);
 
             String accessType = "";
-            if (!grants.isEmpty()) {
+            boolean isBucketGrantExist = true;
+            if (sessionIdRequired) {
+                if (user == null) {
+                    throw new AccessDeniedException("Session id is not active. Please login.");
+                }
```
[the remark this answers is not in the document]
marounkoussaifi: You are right, I'll remove it. Thanks!

src/main/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectGrantController.java
marounkoussaifi: Ok I'll do it.

src/main/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectController.java
```diff
@@ -426,7 +426,19 @@ public CatalogObjectDependencies getDependencies(
...
+                List<BucketGrantMetadata> bucketGrantsMetadata = new LinkedList<>(bucketGrantService.getAllAssignedGrantsForUserAndHisGroups(user.getName(),
+                                                                                                                                             user.getGroups()));
+                bucketGrantsMetadata.removeIf(grant -> !grant.getBucketName().equals(bucketName));
+                if (bucketGrantsMetadata.isEmpty()) {
+                    isBucketGrantExist = false;
+                }
```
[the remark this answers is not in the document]
marounkoussaifi: Ok I'll change it.

fviale (approved):
LGTM

src/main/java/org/ow2/proactive/catalog/service/CatalogObjectGrantService.java
```diff
@@ -338,14 +347,19 @@ public CatalogObjectGrantMetadata updateCatalogObjectGrantForAGroup(String userG
...
+                                                                                                                    catalogObjectName);
+        if (object != null) {
+            long catalogObjectId = object.getId();
+            return catalogObjectGrantRepository.findCatalogObjectGrantEntitiesByBucketEntityIdAndCatalogObjectRevisionEntityId(bucketId,
+                                                                                                                               catalogObjectId)
+                                               .stream()
+                                               .map(CatalogObjectGrantMetadata::new)
+                                               .collect(Collectors.toList());
+        } else {
+            throw new DataIntegrityViolationException("Object: " + catalogObjectName + " was not found in bucket: " +
```
AlexFalessi, line 360: In my opinion this exception should not be manually thrown:

- This is a Spring exception thrown when querying a DB
- The name does not represent the problem
- It will always return a http 500

src/main/java/org/ow2/proactive/catalog/service/BucketService.java
```diff
@@ -217,7 +217,7 @@ public BucketMetadata getBucketMetadata(String bucketName) {
             entities = generateBucketMetadataList(bucketRepository.findAll(sortById));
         }
 
-        log.info("Buckets count {}", entities.size());
+        //log.info("Buckets count {}", entities.size());
```
AlexFalessi, line 220: Uncomment or remove this log please.

AlexFalessi (approved):

src/main/java/org/ow2/proactive/catalog/service/BucketService.java
marounkoussaifi: Ok I'll remove it.

PR #260, opened 2022-02-22
Encrypt PA:HIDDEN variable if not already encrypted

fviale:


src/main/java/org/ow2/proactive/catalog/util/parser/WorkflowParser.java
```diff
@@ -170,6 +173,15 @@ private void addVariableIfNotNullAndModelIfNotEmpty(Set<KeyValueLabelMetadataEnt
         String group = jobVariable.getGroup();
         boolean advanced = jobVariable.isAdvanced();
         boolean hidden = jobVariable.isHidden();
+        if (model != null && "PA:HIDDEN".equalsIgnoreCase(model)) {
+            if (!Strings.isNullOrEmpty(value) && !value.startsWith("ENC(")) {
```
mboussaa, line 177: I prefer to use same functions to check null or empty 

u can replace !Strings.isNullOrEmpty(value) by checkIfNotNull(name, model) && checkIfNotEmpty(name, model)
and
model != null by checkIfNotNull(model)

mboussaa (approved):

src/main/java/org/ow2/proactive/catalog/util/parser/WorkflowParser.java
fviale: what's the point ? you want me to create a function rather than use an existing one ?

mboussaa: checkIfNotNull and checkIfNotEmpty functions already exist

fviale: ah ok, I saw the existing function used already in the class

mboussaa:
ok for me

PR #261, opened 2022-02-23
Fix resulting rights when listing buckets

marounkoussaifi:


marounkoussaifi:
I'm very sorry for not paying attention to this detail yesterday. I only tested for the get Bucket method thinking that it was the only method used to determine the rights for the bucket in the JSON response and I totally forgot about the list bucket method. I retested now for a user that belongs to scheduleradmins and user groups where he has on his user group a grant with a priority higher than 5 and it worked. The show bucket grants button is now hidden and inaccessible for him.

tao-xinxiu (approved):

mklkun (approved):

fviale (approved):
LGTM

PR #266, opened 2022-04-22
Fix the issue regarding a bucket with noAccess grant

marounkoussaifi:
The issue that is solved in this PR was that when there is a bucket grant for a user with noAccess rights, all other buckets' objects are not shown.


fviale (approved):

tao-xinxiu (approved):
Great work. Thank you.

PR #268, opened 2022-05-08
Improve the performance of list catalog objects in a bucket request by reducing the redundant DB queries

tao-xinxiu:
This PR aims to improve the performance of catalog requests of listing buckets and catalog objects:
- GET buckets/ 
- GET buckets/BUCKET_NAME/resources/

Performance test data on a relatively slow PC (local requests):
- GET buckets/: **before** 2000 ms - 3000 ms **after** 800 ms - 1200 ms
- GET buckets/BUCKET_NAME/resources/: **before** 5000 ms - 6000 ms **after** 400 ms - 800 ms 

Important changes:
- avoid redundant DB call: retrieve all the needed data at once (or with fewer DB queries), then filter it to get the needed data. At least, it should be avoided to send DB queries in a loop.
- reduce unnecessary loop
- use Set instead of List when a quick match needed

Small changes:
- reduce duplicate coding
- remove not used functions
- reduce double negative coding
- reduce embedded brackets levels

Potential bugs to be fixed in the future:
- The function `BucketGrantService.getTheNumberOfAccessibleObjectsInTheBucket` and `GrantRightsService.removeAllObjectsWithNoAccessGrant` need to compare grants priority. The current version may cause bugs when the user has multiple conflict group grants (or user grants) over the same catalog object / bucket.

### DIFF
--- a/src/main/java/org/ow2/proactive/catalog/rest/controller/BucketController.java
+++ b/src/main/java/org/ow2/proactive/catalog/rest/controller/BucketController.java
@@ -32,7 +32,6 @@ import static org.springframework.web.bind.annotation.RequestMethod.GET;
 import static org.springframework.web.bind.annotation.RequestMethod.POST;
 import static org.springframework.web.bind.annotation.RequestMethod.PUT;
 
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
@@ -198,7 +197,7 @@ public class BucketController {
             @ApiParam(value = "The name of objects that buckets must contain") @RequestParam(value = "objectName", required = false) Optional<String> objectName)
             throws NotAuthenticatedException, AccessDeniedException {
 
-        List<BucketMetadata> listBucket = new ArrayList<>();
+        List<BucketMetadata> listBucket;
         log.debug("====== Get buckets list request started ======== ");
         long startTime = System.currentTimeMillis();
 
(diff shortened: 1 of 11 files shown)


src/main/java/org/ow2/proactive/catalog/rest/controller/BucketController.java
```diff
@@ -205,31 +206,39 @@ public List<BucketMetadata> list(
...
-                metadata.setRights(bucketGrantAccessType);
-                if (!res.contains(metadata)) {
-                    res.add(metadata);
+
+            List<BucketGrantMetadata> allBucketsGrants = bucketGrantService.getUserAllBucketsGrants(user);
+            List<CatalogObjectGrantMetadata> allCatalogObjectsGrants = catalogObjectGrantService.getAllObjectsGrantsAssignedToAUser(user);
+
+            for (BucketMetadata bucket : listBucket) {
+                if (grantRightsService.isPublicBucket(bucket.getOwner())) {
+                    bucket.setRights(admin.name());
```
marounkoussaifi, line 221: I'm not sure if you should add the bucket object count here also (unless it's added by default when you get the bucket)

marounkoussaifi (approved):

marounkoussaifi:
LGTM! Thank you @tao-xinxiu for the improvements. We need to run manual tests once this PR is merged to guarantee that the changes did not affect the normal behavior.

src/main/java/org/ow2/proactive/catalog/rest/controller/BucketController.java
tao-xinxiu: Yes, it's already added when get the bucket.

fviale (approved):

PR #269, opened 2022-05-13
Add authorisations' modification history

marounkoussaifi:


src/main/java/org/ow2/proactive/catalog/service/BucketGrantService.java
```diff
@@ -118,8 +120,19 @@ public BucketGrantMetadata updateBucketGrantForASpecificUser(String bucketName,
...
             BucketGrantEntity bucketGrantEntity = bucketGrantRepository.findBucketGrantByUsername(bucketId, username);
             if (bucketGrantEntity != null) {
+                String oldValue = bucketGrantEntity.toString();
                 // Update the access type
                 bucketGrantEntity.setAccessType(accessType);
+                // Add modification history
+                ModificationHistoryData modificationHistoryData = new ModificationHistoryData(System.currentTimeMillis(),
+                                                                                              currentUser.getName());
+                // Set old values in history
+                modificationHistoryData.setOldValues(oldValue);
```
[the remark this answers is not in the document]
marounkoussaifi: I did it like this to reduce DB size. Since It's not updated yet, I found it less helpful to add the history since It already has the old values in the grant attributes. When the user edits it, I will save the old ones to show to what they have changed.

fviale:
I don't see an endpoint to retrieve the history for a grant

marounkoussaifi:
@fviale The history of the grant is already obtained by the GET or GET all grants endpoints.

src/main/java/org/ow2/proactive/catalog/service/BucketGrantService.java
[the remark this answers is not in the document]
fviale: ok I get it

fviale:
@marounkoussaifi ok

fviale (approved):

tao-xinxiu (approved):

src/main/java/org/ow2/proactive/catalog/dto/BucketGrantMetadata.java
```diff
@@ -25,7 +25,10 @@
  */
 package org.ow2.proactive.catalog.dto;
 
+import java.util.Deque;
```
mklkun, line 28: Why are you using `Deque` why not `List`?

marounkoussaifi: We want a LIFO order. A list will provide the reverse.

mklkun: Fair enough. LGTM

PR #271, opened 2022-05-23
have-alphabetical-order-of-buckets

oanaBiancaSchiopu:


src/main/java/org/ow2/proactive/catalog/dto/CatalogObjectNameReference.java
```diff
@@ -51,4 +52,15 @@ public CatalogObjectNameReference(String bucketName, String projectName, String
         this.projectName = projectName;
         this.name = name;
     }
+
+    @Override
+    public int compareTo(@NonNull CatalogObjectNameReference catalogObjectNameReference) {
+        if (getBucketName() == null || catalogObjectNameReference.getBucketName() == null || getName() == null ||
+            catalogObjectNameReference.getName() == null) {
```
aminelouati, line 59: Is it possible that getName can be null ?

oanaBiancaSchiopu: Without this annotation  I receive a warning saying that "Not annotated parameter overrides @NotNull parameter". I guess the parent method has some @NotNull annotation on the parameter so I had to add it to the overriding method

oanaBiancaSchiopu: sorry, I understand wrong your question. I guess, getName cannot be null. You think that this check is useless?

aminelouati (approved):

PR #274, opened 2022-06-15
Fix: calculate grants priority when decides the accessible objects in bucket

tao-xinxiu:
- In the func `getNumberOfAccessibleObjectsInBucket` and `removeInaccessibleObjectsInBucket`, its logic should based on the grant rule that userObject grant > userBucket grant > groupObject grant > groupBucket grant.
- refactor the class GrantRightsService, BucketGrantService, and CatalogObjectGrantService: 
  - Add common parent class GrantMetadata for BucketGrantMetadata and CatalogObjectGrantMetadata to reduce duplicated code 
  - Add util class GrantHelper to make the code more concise in the big classes: GrantRightsService, BucketGrantService and CatalogObjectGrantService.
  - Change the functions order in the class GrantRightsService, BucketGrantService, and CatalogObjectGrantService to make functions easier to compare and find the proper one to use.
  - Use a more common practice way to compare access type.


src/main/java/org/ow2/proactive/catalog/repository/BucketGrantRepository.java
```diff
@@ -54,27 +54,21 @@ public interface BucketGrantRepository extends JpaRepository<BucketGrantEntity,
...
-            long bucketId);
+    List<BucketGrantEntity> findAccessibleBucketsGrantsAssignedToUserGroups(List<String> userGroup, long bucketId);
 
-    @Query(value = "SELECT bge FROM BucketGrantEntity bge WHERE bge.accessType='noAccess'")
-    List<BucketGrantEntity> findAllBucketGrantsWithNoAccessRight();
+    @Query(value = "SELECT bge FROM BucketGrantEntity bge WHERE bge.grantee = ?1 And bge.granteeType='user' AND bge.accessType='noAccess'")
+    List<BucketGrantEntity> findBucketsGrantsAssignedToAUsernameWithNoAccessRight(String username);
 
-    @Query(value = "SELECT bge FROM BucketGrantEntity bge WHERE bge.grantee = ?1 AND bge.bucketEntity.id=?2 And bge.granteeType='user' AND bge.accessType='noAccess'")
-    List<BucketGrantEntity> findBucketGrantsAssignedToAUsernameWithNoAccessRight(String username, long bucketId);
```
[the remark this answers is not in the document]
tao-xinxiu: Yes, actually they were not used even before. I guess it's because of code changes.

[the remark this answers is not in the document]
tao-xinxiu: The case of when needing noAccess grants it's better to be filtered out from all user bucket grants instead of separately doing DB requests each time.

src/main/java/org/ow2/proactive/catalog/util/GrantHelper.java
```diff
@@ -0,0 +1,193 @@
...
+     *
+     * @param bucketOwner bucket owner (complete format, "GROUP:XXXXX")
+     * @return the group name that the bucket belongs to
+     */
+    public static String extractBucketOwnerGroup(String bucketOwner) {
+        return bucketOwner.substring(BUCKET_OWNER_PREFIX_LENGTH);
+    }
+
+    public static BucketGrantMetadata ownerGroupGrant(String bucketOwner, String bucketName) {
+        return new BucketGrantMetadata("group", "", bucketOwner.substring(6), "admin", 5, 0, bucketName);
```
fviale, line 68: if "group:" or "group" is defined somewhere as a constant, it would be cleaner to use

substring(GROUP_PREFIX.length())

src/main/java/org/ow2/proactive/catalog/util/GrantHelper.java
```diff
@@ -0,0 +1,193 @@
...
+                     .collect(Collectors.toList());
+    }
+
+    public static <T extends GrantMetadata> List<T> filterGrantsAssignedToUser(AuthenticatedUser user, List<T> grants) {
+        return grants.stream().filter(grant -> isGrantAssignedToUser(user, grant)).collect(Collectors.toList());
+    }
+
+    public static boolean isGrantAssignedToUser(AuthenticatedUser user, GrantMetadata grant) {
+        return (grant.getGrantee().equals(user.getName()) && grant.getGranteeType().equals("user")) ||
+               (user.getGroups().contains(grant.getGrantee()) && grant.getGranteeType().equals("group"));
```
fviale, line 173: similarly you can introduce here boolean grantAppliesToAGroupOfThisUser(...)

src/main/java/org/ow2/proactive/catalog/util/GrantHelper.java
```diff
@@ -0,0 +1,193 @@
...
+
+public class GrantHelper {
+    private static final String GROUP_GRANTEE_TYPE = "group";
+
+    private static final String USER_GRANTEE_TYPE = "user";
+
+    // bucket owner is in the format of "GROUP:XXXXX"
+    private static final int BUCKET_OWNER_PREFIX_LENGTH = OwnerGroupStringHelper.GROUP_PREFIX.length();
+
+    public static boolean isPublicBucket(String bucketOwner) {
```
fviale, line 53: shouldn't this method be

`boolean isPublicBucket(BucketMetadata bucket)` ?

src/main/java/org/ow2/proactive/catalog/util/GrantHelper.java
```diff
@@ -0,0 +1,193 @@
...
+import org.ow2.proactive.catalog.dto.GrantMetadata;
+import org.ow2.proactive.catalog.repository.entity.BucketGrantEntity;
+import org.ow2.proactive.catalog.repository.entity.CatalogObjectGrantEntity;
+import org.ow2.proactive.catalog.service.BucketService;
+import org.ow2.proactive.catalog.service.OwnerGroupStringHelper;
+import org.ow2.proactive.catalog.service.model.AuthenticatedUser;
+
+
+public class GrantHelper {
+    private static final String GROUP_GRANTEE_TYPE = "group";
```
fviale, line 46: these constant should be defined public static final and used in place of "group" or "user" strings everywhere in the catalog project.

Example, it is not used in ownerGroupGrant method

src/main/java/org/ow2/proactive/catalog/util/GrantHelper.java
tao-xinxiu: It's used both for BucketMetadata and BucketEntity. But yes, I could change it.

src/main/java/org/ow2/proactive/catalog/util/GrantHelper.java
```diff
@@ -0,0 +1,193 @@
...
+                     .map(CatalogObjectGrantMetadata::new)
+                     .collect(Collectors.toList());
+    }
+
+    public static <T extends GrantMetadata> List<T> filterGrantsAssignedToUser(AuthenticatedUser user, List<T> grants) {
+        return grants.stream().filter(grant -> isGrantAssignedToUser(user, grant)).collect(Collectors.toList());
+    }
+
+    public static boolean isGrantAssignedToUser(AuthenticatedUser user, GrantMetadata grant) {
+        return (grant.getGrantee().equals(user.getName()) && grant.getGranteeType().equals("user")) ||
```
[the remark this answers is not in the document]
tao-xinxiu: These three func is for different types of objects: GrantMetadata, BucketGrantEntity and CatalogObjectGrantEntity. But yes, you're right, I'll create a separated func with parameters granteeType and grantee to reuse it.

PR #275, opened 2022-06-23
Fix duplicated buckets with grants and disallow downgrade user's admin rights

tao-xinxiu:
Fixed two bugs: 

- We should not show duplicated buckets when the user is both in the bucket owner group and has grants on the bucket. 
- When updating a group grant, even when it's just to change the grant priority, we should never allow a user to downgrade its admin right over the bucket. Otherwise, the bucket owner may lost its admin rights through first create a lower-priority not-admin grant, then increase the grant priority to be larger than 5.


nebilbenmabrouk (approved):
LGTM

aminelouati (approved):

PR #277, opened 2022-07-05
Filter buckets by name in the Catalog portal

amouhoub:


src/main/java/org/ow2/proactive/catalog/rest/controller/BucketController.java
```diff
@@ -183,8 +184,10 @@ public List<BucketMetadata> list(
             @ApiParam(value = "The name of the user who owns the Bucket") @RequestParam(value = "owner", required = false) String ownerName,
             @ApiParam(value = "The kind(s) of objects that buckets must contain. Multiple kinds can be specified using comma separators") @RequestParam(value = "kind", required = false) Optional<String> kind,
             @ApiParam(value = "The Content-Type of objects that buckets must contain") @RequestParam(value = "contentType", required = false) Optional<String> contentType,
-            @ApiParam(value = "The name of objects that buckets must contain") @RequestParam(value = "objectName", required = false) Optional<String> objectName)
-            throws NotAuthenticatedException, AccessDeniedException {
+            @ApiParam(value = "The name of objects that buckets must contain") @RequestParam(value = "objectName", required = false) Optional<String> objectName,
+            @ApiParam(value = "Starts with bucket name (case insensitive)")
```
fviale, line 188: after double checking, search by object names uses contains and not startsWith, so I think we should do the same for bucketName

src/main/java/org/ow2/proactive/catalog/rest/controller/BucketController.java
```diff
@@ -194,6 +197,7 @@ public List<BucketMetadata> list(
         kind = kind.filter(s -> !s.isEmpty());
         contentType = contentType.filter(s -> !s.isEmpty());
         objectName = objectName.filter(s -> !s.isEmpty());
+        bucketName.filter(s -> !s.isEmpty());
```
fviale, line 200: you need to write

bucketName = bucketName.filter(s -> !s.isEmpty());

amouhoub: The bucketName should be declared as `final` to avoid issues in the lambda.
I'll create a new final temp variable for this

PR #286, opened 2022-11-19
Improve search by tag performance

fviale:
Use database search instead of filtering hibernate entities


tao-xinxiu (approved):
LGTM. Great work.

aminelouati (approved):

PR #291, opened 2022-12-23
Handle % character when used in CatalogObject by user

mklkun:


src/main/java/org/ow2/proactive/catalog/repository/CatalogObjectRevisionRepositoryImpl.java
```diff
@@ -75,6 +75,16 @@ public List<CatalogObjectRevisionEntity> findDefaultCatalogObjectsOfKindListAndC
                  .collect(Collectors.toList());
     }
 
+    private String toRightSidePredicatePattern(String pattern) {
+        return pattern.endsWith("%") ? pattern.toLowerCase() : pattern.toLowerCase() + "%";
+    }
+
+    private String toBothSidesPredicatePattern(String pattern) {
+        String predicatePattern = toRightSidePredicatePattern(pattern);
+        predicatePattern = predicatePattern.startsWith("%") ? predicatePattern : "%" + predicatePattern;
```
fviale, line 84: this does not seem to allow use input with "word%" as a percentage will always be added in the beginning.

fviale: the idea is more "if the pattern contains % don't change it, otherwise add default patterns"

src/main/java/org/ow2/proactive/catalog/repository/CatalogObjectRevisionRepositoryImpl.java
```diff
@@ -75,6 +75,14 @@ public List<CatalogObjectRevisionEntity> findDefaultCatalogObjectsOfKindListAndC
                  .collect(Collectors.toList());
     }
 
+    private String toRightSidePredicatePattern(String pattern) {
+        return pattern.contains("%") ? pattern.toLowerCase() : pattern.toLowerCase() + "%";
+    }
+
+    private String toBothSidesPredicatePattern(String pattern) {
+        return pattern.contains("%") ? pattern.toLowerCase() : "%" + pattern + "%";
```
AlexFalessi, line 83: Missing `.toLowerCase()` for false case

mklkun: You're right! I forgot it. It will be fixed. Thanks!

fviale (approved):

AlexFalessi (approved):

PR #292, opened 2023-01-16
Add the possibility to create/update tags for any Catalog Object

aminelouati:


src/main/java/org/ow2/proactive/catalog/dto/CatalogObjectMetadata.java
```diff
@@ -141,17 +143,21 @@ public CatalogObjectMetadata(String bucketName, String name, String projectName,
...
-        Optional<Metadata> projectNameIfExists = metadataList.stream()
-                                                             .filter(property -> property.getKey()
-                                                                                         .equals("project_name"))
-                                                             .findAny();
-        return projectNameIfExists.map(value -> value.getValue() == null ? "" : value.getValue()).orElse("");
+    private String getMetadataValueIfExistsOrEmptyString(String metadataKey) {
+        Optional<Metadata> metadataIfExists = metadataList.stream()
+                                                          .filter(property -> property.getKey().equals(metadataKey))
+                                                          .findAny();
+        return metadataIfExists.map(value -> value.getValue() == null ? "" : value.getValue()).orElse("");
```
AlexFalessi, line 160: I think it's better to put the second test (`value.getValue() == null ?`) in the filter.
Also you can directly return.

Something like

```
return metadataList.stream()
                .filter(property -> property.getKey().equals(metadataKey) && !StringUtils.isBlank(property.getValue()))
                .findFirst()
                .map(Metadata::getValue)
                .orElse("");
```

src/main/java/org/ow2/proactive/catalog/service/CatalogObjectService.java
```diff
@@ -620,26 +694,39 @@ public CatalogObjectMetadata createCatalogObjectRevision(String bucketName, Stri
...
-        return projectNameIfExists.map(value -> value.getValue() == null ? "" : value.getValue()).orElse("");
+    protected String getTagsValueIfExistsOrEmptyString(List<Metadata> metadataListParsed, String metadataLabel) {
+        List<String> tagsValueIfExists = metadataListParsed.stream()
+                                                           .filter(property -> property.getLabel()
+                                                                                       .equals(metadataLabel))
+                                                           .collect(Collectors.toList())
+                                                           .stream()
+                                                           .map(Metadata::getKey)
+                                                           .collect(Collectors.toList());
+        return (tagsValueIfExists.isEmpty() ? "" : String.join(",", tagsValueIfExists));
```
AlexFalessi, line 713: I think you can simplify doing directly

```
return metadataListParsed.stream()
                .filter(property -> property.getLabel()
                        .equals(metadataLabel))
                .map(Metadata::getKey)
                .collect(Collectors.joining(","));
```

` .collect(Collectors.joining(","));` returns an empty string if the stream is empty

src/main/java/org/ow2/proactive/catalog/service/CatalogObjectService.java
```diff
@@ -620,26 +694,39 @@ public CatalogObjectMetadata createCatalogObjectRevision(String bucketName, Stri
...
+    }
+
+    protected String getProjectNameIfExistsOrEmptyString(List<Metadata> metadataListParsed, String projectNameKey) {
+        Optional<Metadata> metadataIfExists = metadataListParsed.stream()
+                                                                .filter(property -> property.getKey()
+                                                                                            .equals(projectNameKey))
+                                                                .findAny();
+        return metadataIfExists.map(value -> value.getValue() == null ? "" : value.getValue()).orElse("");
```
AlexFalessi, line 721: Same here with

```
return metadataListParsed.stream()
                .filter(property -> property.getKey()
                        .equals(projectNameKey) && !StringUtils.isBlank(property.getValue()))
                .findAny()
                .map(Metadata::getValue)
                .orElse("");
```

AlexFalessi (approved):
Small non blocking comments. Nice PR !

oanaBiancaSchiopu (approved):

fviale:
Retest this please

fviale (approved):
LGTM

PR #294, opened 2023-03-12
fix updateObjectMetadata service

aminelouati:


src/main/java/org/ow2/proactive/catalog/service/CatalogObjectService.java
```diff
@@ -456,6 +439,11 @@ private CatalogObjectRevisionEntity buildCatalogObjectRevisionEntity(final Strin
         //the tag comma separated string is synchronized in the following order tags > metadataTags > workflowXmlTags
         String synchronizedTags = synchronizeMetadataValue(tags, metadataTags, workflowXmlTags);
 
+        if (commitMessage.equals("The project name or/and tags metadata are updated")) {
```
fviale, line 442: ouch, is there a better way to do this ?

src/test/java/org/ow2/proactive/catalog/service/CatalogObjectServiceTest.java
```diff
@@ -397,10 +406,11 @@ public void testUpdateObjectMetadataOnlyKindOrOnlyContentTypeOrOnlyProjectName()
                                                                                                           Optional.of("updated-kind"),
                                                                                                           Optional.of("updated-contentType"),
                                                                                                           Optional.of("updated-projectName"),
-                                                                                                          Optional.of(TAGS));
+                                                                                                          Optional.of(TAGS),
+                                                                                                          USERNAME);
 
         assertThat(catalogObjectUpdatedProjectName).isNotNull();
-        assertThat(catalogObjectUpdatedProjectName.getCommitMessage()).isEqualTo(COMMIT_MESSAGE);
+        assertThat(catalogObjectUpdatedProjectName.getCommitMessage()).isEqualTo("The project name or/and tags metadata are updated");
```
fviale, line 413: if this commit message is used in multiple place, consider defining a constant

src/main/java/org/ow2/proactive/catalog/service/CatalogObjectService.java
```diff
@@ -277,31 +277,16 @@ public CatalogObjectMetadata updateObjectMetadata(String bucketName, String name
...
-        for (KeyValueLabelMetadataEntity metadata : catalogObjectRevisionEntity.getKeyValueMetadataList()) {
-            if (metadata.getKey().equals(PROJECT_NAME)) {
-                metadata.setValue(projectName.orElse(""));
-            }
+        if (!catalogObjectRevisionEntity.getProjectName().equals(projectName.orElse("")) ||
+            !catalogObjectRevisionEntity.getTags().equals(tags.orElse(""))) {
+            catalogObjectRevisionEntity = buildCatalogObjectRevisionEntity("The project name or/and tags metadata are updated",
+                                                                           username,
+                                                                           projectName.orElse(""),
+                                                                           tags.orElse(""),
```
fviale, line 285: will the portal provide existing tags when updating metadata ?

If not, this change means that when a user tries to update only the projectName, the code will set at the same time  tags to "".

src/main/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectController.java
```diff
@@ -231,21 +231,30 @@ public CatalogObjectMetadata updateObjectMetadata(
...
 
             // Check Grants
-            AuthenticatedUser user = restApiAccessService.getUserFromSessionId(sessionId);
+            user = restApiAccessService.getUserFromSessionId(sessionId);
             if (!AccessTypeHelper.satisfy(grantRightsService.getCatalogObjectRights(user, bucketName, name), write)) {
                 throw new CatalogObjectGrantAccessException(bucketName, name);
             }
 
+        } else {
+            user = AuthenticatedUser.EMPTY;
```
fviale, line 248: in which case would this happen ?

src/main/java/org/ow2/proactive/catalog/service/CatalogObjectService.java
```diff
@@ -277,31 +277,16 @@ public CatalogObjectMetadata updateObjectMetadata(String bucketName, String name
...
         }
 
-        //synchronize project name values
-        for (KeyValueLabelMetadataEntity metadata : catalogObjectRevisionEntity.getKeyValueMetadataList()) {
-            if (metadata.getKey().equals(PROJECT_NAME)) {
-                metadata.setValue(projectName.orElse(""));
-            }
+        if (!catalogObjectRevisionEntity.getProjectName().equals(projectName.orElse("")) ||
+            !catalogObjectRevisionEntity.getTags().equals(tags.orElse(""))) {
+            catalogObjectRevisionEntity = buildCatalogObjectRevisionEntity("The project name or/and tags metadata are updated",
```
fviale, line 282: better call this catalogObjectNewRevisionEntity

src/main/java/org/ow2/proactive/catalog/service/CatalogObjectService.java
aminelouati: I know it is dirty... I can add an additional parameter (boolean) to consider the update situation.

fviale: yeah, that would be preferable :-)

src/main/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectController.java
aminelouati: Used to initialize the user.name with an empty string

src/main/java/org/ow2/proactive/catalog/service/CatalogObjectService.java
aminelouati: Yes, the portal provides existing tags when updating metadata.

src/test/java/org/ow2/proactive/catalog/service/CatalogObjectServiceTest.java
aminelouati: OK

src/main/java/org/ow2/proactive/catalog/service/CatalogObjectService.java
aminelouati: The variable assignment is useless. I removed it

src/test/java/org/ow2/proactive/catalog/service/CatalogObjectServiceTest.java
```diff
@@ -72,6 +72,8 @@ public class CatalogObjectServiceTest {
 
     public static final String COMMIT_MESSAGE = "commit message";
 
+    public static final String UPDATE_COMMIT_MESSAGE = "The project name or/and tags metadata are updated";
```
fviale, line 75: why define this constant in the Test class ? you are using the message in the main code also

src/integration-test/java/org/ow2/proactive/catalog/service/CatalogObjectServiceIntegrationTest.java
```diff
@@ -1264,8 +1307,9 @@ public void testUpdateObjectMetadata() {
                                                                                                 Optional.of("updated-kind"),
                                                                                                 Optional.of("updated-contentType"),
                                                                                                 Optional.of("updated-projectName"),
-                                                                                                Optional.of("tag1,tag2,tag3"));
-        assertThat(catalogObjectMetadata.getCommitMessage()).isEqualTo("commit message 2");
+                                                                                                Optional.of("tag1,tag2,tag3"),
+                                                                                                "username");
+        assertThat(catalogObjectMetadata.getCommitMessage()).isEqualTo("The project name or/and tags metadata are updated");
```
fviale, line 1312: please use the constant

src/main/java/org/ow2/proactive/catalog/service/CatalogObjectService.java
```diff
@@ -277,31 +278,17 @@ public CatalogObjectMetadata updateObjectMetadata(String bucketName, String name
...
         }
 
-        //synchronize project name values
-        for (KeyValueLabelMetadataEntity metadata : catalogObjectRevisionEntity.getKeyValueMetadataList()) {
-            if (metadata.getKey().equals(PROJECT_NAME)) {
-                metadata.setValue(projectName.orElse(""));
-            }
+        if (!catalogObjectRevisionEntity.getProjectName().equals(projectName.orElse("")) ||
+            !catalogObjectRevisionEntity.getTags().equals(tags.orElse(""))) {
+            buildCatalogObjectRevisionEntity("The project name or/and tags metadata are updated",
```
fviale, line 283: constant ?

src/main/java/org/ow2/proactive/catalog/service/CatalogObjectService.java
```diff
@@ -277,31 +278,17 @@ public CatalogObjectMetadata updateObjectMetadata(String bucketName, String name
...
+                                             username,
+                                             projectName.orElse(""),
+                                             tags.orElse(""),
```
fviale, line 286: I would prefer that, if the projectName or tags is absent, we take the value from the previous revision instead of setting it to empty.

src/test/java/org/ow2/proactive/catalog/service/CatalogObjectServiceTest.java
fviale: you should import the constant instead of redefining it

src/integration-test/java/org/ow2/proactive/catalog/service/CatalogObjectServiceIntegrationTest.java
```diff
@@ -92,6 +92,8 @@ public class CatalogObjectServiceIntegrationTest {
 
     private static final String PROJECT_NAME = "projectName";
 
+    public static final String UPDATE_COMMIT_MESSAGE = "The project name or/and tags metadata are updated";
```
fviale, line 95: you should import the constant instead of redefining it

PR #301, opened 2023-06-22
Log every user action in the catalog

fviale:
Also:
  - ensure that authentication and RBAC is enforced in delete empty buckets
  - fix tests in fast machines (add tiny waits to avoid 2 revisions with the same commit time in ms)

### DIFF
--- a/src/test/java/org/ow2/proactive/catalog/service/CatalogObjectServiceTest.java
+++ b/src/test/java/org/ow2/proactive/catalog/service/CatalogObjectServiceTest.java
@@ -312,7 +312,7 @@ public class CatalogObjectServiceTest {
     }
 
     @Test
-    public void testUpdateObjectMetadata() {
+    public void testUpdateObjectMetadata() throws InterruptedException {
         long now = System.currentTimeMillis();
         BucketEntity bucketEntity = new BucketEntity("bucket", "toto");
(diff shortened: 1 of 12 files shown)


src/test/java/org/ow2/proactive/catalog/service/CatalogObjectServiceTest.java
```diff
@@ -424,6 +436,9 @@ public void testUpdateObjectMetadataOnlyKindOrOnlyContentTypeOrOnlyProjectNameOr
         assertThat(catalogObjectUpdatedProjectName.getMetadataList()).hasSize(2);
         assertThat(catalogObjectUpdatedProjectName.getCommitTimeRaw()).isNotEqualTo(String.valueOf(now));
 
+        // a necessary sleep to avoid that the new revision share the same commit time (and thus is discarded)
+        Thread.sleep(2);
```
mbenguig, line 440: tinyWait is not accessible from there ?

mbenguig (approved):
A very  "tiny" comment.

src/test/java/org/ow2/proactive/catalog/service/CatalogObjectServiceTest.java
fviale: I can create a method yes

fviale:
Retest this please

AlexFalessi (approved):

PR #302, opened 2023-07-21
Add filtering according to job-planner association status

fviale:
- added a JobPlannerService (with configuration and cache of 1 minute)
 - add new associationStatus filter in BucketController and CatalogObjectController
 - association status information is returned as a metadata of the catalog objetcs

### DIFF
--- a/src/main/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectController.java
+++ b/src/main/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectController.java
@@ -25,6 +25,7 @@
  */
 package org.ow2.proactive.catalog.rest.controller;
 
+import static org.ow2.proactive.catalog.dto.AssociationStatus.UNPLANNED;
 import static org.ow2.proactive.catalog.util.AccessType.*;
 import static org.springframework.web.bind.annotation.RequestMethod.DELETE;
 import static org.springframework.web.bind.annotation.RequestMethod.GET;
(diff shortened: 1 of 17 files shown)


AlexFalessi (approved):

andrewssobral (approved):

src/main/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectController.java
```diff
@@ -384,6 +397,7 @@ public ResponseEntity<List<CatalogObjectMetadata>> list(
             @ApiParam(value = "Filter according to Content-Type.") @RequestParam(required = false) Optional<String> contentType,
             @ApiParam(value = "Filter according to Object Name.") @RequestParam(value = "objectName", required = false) Optional<String> objectNameFilter,
             @ApiParam(value = "Filter according to Object Tag.") @RequestParam(value = "objectTag", required = false) Optional<String> objectTagFilter,
+            @ApiParam(value = "Filter according to Job-Planner association status. If enabled, only objects for which a job-planner association exists with the provided status will be returned. Can be PLANNED, DEACTIVATED, FAILED or UNPLANNED") @RequestParam(value = "associationStatus", required = false) Optional<String> associationStatusFilter,
```
amouhoub, line 400: @fviale According to Denis, we need a :"All" value for all 3 planning states. (Unplanned is for objects not being associated in JP at all.
Is this all state inferred from the absence of this query parameter ?

fviale: @amouhoub the All value is exactly as you said, just don't add the filter.

fviale: but yes, I did not implement a All that would only select objects with either of PLANNED, DEACTIVATED or FAILED. I did not know this was requested.

PR #304, opened 2023-07-24
Allow pattern filtering on BucketController.list

fviale:
Previously, pattern filtering (for objectName, contentType or kind) was implemented only in CatalogObjectController.list


mboussaa (approved):
LGTM

PR #312, opened 2023-09-19
Add filters

AlexFalessi:
Add project name, last commited by and time range filters to GET buckets and GET catalog object metadata


oanaBiancaSchiopu (approved):

src/integration-test/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectReportControllerIntegrationTest.java
```diff
@@ -85,6 +85,7 @@ public void setup() throws IOException {
                .queryParam("name", bucketObjectName)
                .queryParam("commitMessage", "commit message")
                .queryParam("objectContentType", MediaType.APPLICATION_XML.toString())
+               .queryParam("projectName", MediaType.APPLICATION_XML.toString())
```
fviale, line 88: MediaType.APPLICATION_XML.toString() is not the default value (or example usage) of projectName

src/main/java/org/ow2/proactive/catalog/rest/controller/BucketController.java
```diff
@@ -206,7 +207,7 @@ public List<BucketMetadata> list(
...
+            final Optional<String> bucketName, @ApiParam(value = "The project name equals this parameter (case insensitive)") @RequestParam(value = "projectName", required = false) Optional<String> projectName, @ApiParam(value = "The user who last committed changes in the buckets") @RequestParam(value = "username", required = false) Optional<String> lastCommittedBy, @ApiParam(value = "The time after the last commit took place in the buckets in ms") @RequestParam(value = "committedTimeGreater", required = false, defaultValue = "0") Optional<Long> committedTimeGreater, @ApiParam(value = "The time before the last commit took place in the buckets in ms") @RequestParam(value = "committedTimeLessThan", required = false, defaultValue = "0") Optional<Long> committedTimeLessThan, @ApiParam(value = "If true, buckets without objects matching the filters will be returned with objectCount=0. Default is false") @RequestParam(value = "allBuckets", required = false, defaultValue = "false") String allBuckets) throws NotAuthenticatedException, AccessDeniedException {
```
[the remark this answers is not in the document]
fviale: The user who last committed changes in the buckets
=> this is a strange filter honestly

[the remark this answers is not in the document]
fviale: The time after the last commit took place in the buckets in ms
=> filters only buckets where the last commit occurred after the given EPOCH time.
The time before the last commit took place in the buckets in ms
=> filters only buckets where the last commit occurred before the given EPOCH time (this filter is a bit strange as it will return any bucket where there exists a least one old commit which will likely be "all buckets")

src/main/java/org/ow2/proactive/catalog/rest/controller/CatalogObjectController.java
```diff
@@ -399,6 +399,10 @@ public ResponseEntity<List<CatalogObjectMetadata>> list(
             @ApiParam(value = "Filter according to Object Name.") @RequestParam(value = "objectName", required = false) Optional<String> objectNameFilter,
             @ApiParam(value = "Filter according to Object Tag.") @RequestParam(value = "objectTag", required = false) Optional<String> objectTagFilter,
             @ApiParam(value = "Filter according to Job-Planner association status. If enabled, only objects for which a job-planner association exists with the provided status will be returned. ALL, PLANNED, DEACTIVATED, FAILED or UNPLANNED. ALL will filter objects which have an association with any status. UNPLANNED will filter objects without any association.") @RequestParam(value = "associationStatus", required = false) Optional<String> associationStatusFilter,
+            @ApiParam(value = "Filter according to Project Name.") @RequestParam(value = "projectName", required = false) Optional<String> projectNameFilter,
```
amouhoub, line 402: Filter by ...,
same goes for the other ones above (old ones)

src/main/java/org/ow2/proactive/catalog/repository/BucketRepositoryImpl.java
```diff
@@ -51,28 +51,35 @@ public class BucketRepositoryImpl implements BucketRepositoryCustom {
...
-        return em.createQuery(buildCriteriaQuery(null, kindList, contentType, objectName)).getResultList();
+        return em.createQuery(buildCriteriaQuery(null, kindList, contentType, objectName, null, null)).getResultList();
     }
 
     @Override
-    public List<Object[]> findBucketByOwnerContainingKindListAndContentTypeAndObjectName(List<String> owners,
-            List<String> kindList, String contentType, String objectName) {
-
-        return em.createQuery(buildCriteriaQuery(owners, kindList, contentType, objectName)).getResultList();
+    public List<Object[]> findBucketByOwnerContainingKindListAndContentTypeAndObjectNameAndLastCommittedTimeInterval(
```
amouhoub, line 58: Maybe we should rename this method and avoid redundancy of filtering methods by refactoring them  but I don't know how hard it is to refactor this

- findBucketByOwnerContainingKindListAndContentTypeAndObjectName
- findBucketByOwnerContainingKindList
- findBucketByOwnerContainingKindListAndContentTypeAndObjectNameAndLastCommittedTimeInterval
- etc...

This would have been acceptable in a test code, but since, practically we have a single endpoint that uses multiple filters, its better to use one generic filterBuckets() method with a list of filters or a BucketFilters entity....
what do you think @fviale 

Because otherwise, this code looks very complex and a little confusing for future developers.
Especially that we don't use filters this way in other micro-services and in the scheduler itself

AlexFalessi: It's true for methods in the repositories interfaces too like `BucketRepositoryCustom`

fviale (approved):

src/main/java/org/ow2/proactive/catalog/repository/BucketRepositoryImpl.java
```diff
@@ -112,6 +119,14 @@ private CriteriaQuery<Object[]> buildCriteriaQuery(List<String> owners, List<Str
             Predicate ownerPredicate = cb.in(bucketEntityRoot.get("owner")).value(owners);
             allPredicates.add(ownerPredicate);
         }
+
+        if (committedTimeGreater > 0) {
+            allPredicates.add(cb.ge(catalogObjectsJoin.get("lastCommitTime"), committedTimeGreater));
+        }
+        if (committedTimeLessThan > 0) {
+            allPredicates.add(cb.lessThan(catalogObjectsJoin.get("lastCommitTime"), committedTimeLessThan));
+        }
```
aminelouati, line 128: I believe that you can exclude both filters in case committedTimeLessThan > committedTimeGreater

aminelouati (approved):

src/main/java/org/ow2/proactive/catalog/repository/BucketRepositoryImpl.java
AlexFalessi: You can't do that using the UI. When you select a start date, the dates before are disabled.
If you force it by manually entering a date in the past, once the form is submitted, it will change to the same date as the start date.
Finally, using the REST api, I personally think that it is confusing to just ignore the filter. Results will be returned and the user can think that these results satisfy the filter. As it is now, you will just get an empty response. 
The other way to handle this in my opinion is to throw an exception

AlexFalessi:
Retest this please

PR #317, opened 2023-11-03
Replace springfox swagger with springdoc OpenAPI 3

fviale:
Changed all Swagger annotations to their equivalent in OpenAPI 3 format

Better handle pdf reports (can now be used from the open api portal)

Add some constraints to parameters as the OpenAPI 3 format allows


nebilbenmabrouk (approved):
LGTM

PR #330, opened 2024-01-16
Add import package feature

fviale:
- Refactor existing code to avoid duplicates
- support projectName and tags in export and import
- added an intgration test


amouhoub (approved):
LGTM !

PR #332, opened 2024-03-15
Return empty list if user hasn't got the authorization on a bucket in…

AlexFalessi:
…stead of exception throw when fetching catalog objects


fviale (approved):
ok but this might break some system tests

AlexFalessi:
I have failing tests but I think it's because of the defined user accounts and their rights that are not the same in system-tests-only and local. I'll handle any failing test tomorrow with a new build.

AlexFalessi:
Honestly I don't see a test that should fail because of this changed

oanaBiancaSchiopu (approved):

PR #334, opened 2024-08-14
Fix pdf gen for all buckets

AlexFalessi:
Fix generate PDF report for all Buckets feature by generating a PDF report for each bucket and packing them in a zip file.


AlexFalessi:
Retest this please

PR #335, opened 2024-08-19
Do like catalog object PDF report for all catalog objects = make GET …

AlexFalessi:
…call graph report for all buckets return a ZIP with a PDF report for all buckets.

### DIFF
--- a/src/main/java/org/ow2/proactive/catalog/service/CatalogObjectCallGraphService.java
+++ b/src/main/java/org/ow2/proactive/catalog/service/CatalogObjectCallGraphService.java
@@ -25,14 +25,19 @@
  */
 package org.ow2.proactive.catalog.service;
 
-import java.util.List;
-import java.util.Optional;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.*;
+import java.util.stream.Collectors;
 
 import org.ow2.proactive.catalog.callgraph.CatalogObjectCallGraphPDFGenerator;
 import org.ow2.proactive.catalog.dto.CatalogObjectMetadata;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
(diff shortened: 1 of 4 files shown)


src/main/java/org/ow2/proactive/catalog/service/CatalogObjectCallGraphService.java
```diff
@@ -52,15 +57,33 @@ public class CatalogObjectCallGraphService {
...
+    public byte[] generateBytesCallGraphForAllBucketsAsZip(List<String> authorisedBucketsNames, Optional<String> kind,
             Optional<String> contentType) {
 
-        List<CatalogObjectMetadata> metadataList = catalogObjectService.listCatalogObjects(authorisedBucketsNames,
-                                                                                           kind,
-                                                                                           contentType);
-
-        return catalogObjectCallGraphPDFGenerator.generatePdfImage(metadataList, kind, contentType);
-
+        Map<String, List<CatalogObjectMetadata>> bucketNameCatalogObjectMetadata = catalogObjectService.listCatalogObjects(authorisedBucketsNames,
```
ankicabarisic, line 63: byteArrayOutputStream.toByteArray() call will create a copy of the entire output in memory, which can be problematic if we have very large outputs. Would consider streaming the data or processing the PDFs incrementally.

ankicabarisic (approved):
The code is logically sound and achieves the intended functionality of generating PDF reports from catalog objects and packaging them into a zip file.
Looks good.

src/main/java/org/ow2/proactive/catalog/service/CatalogObjectCallGraphService.java
AlexFalessi: I agree it would be better but my goal wasn't to refactor the PDF generation process but just update the result to fix the feature. I think that creating the PDF incrementally would require quite the refactoring.
The threshold here is more than 700 objects in a single bucket.

PR #340, opened 2024-09-24
Add PUT endpoint to update metadata of multiple catalog objects

AlexFalessi:


oanaBiancaSchiopu (approved):

src/main/java/org/ow2/proactive/catalog/service/CatalogObjectService.java
```diff
@@ -349,26 +360,45 @@ public CatalogObjectMetadata updateObjectMetadata(String bucketName, String name
...
-        catalogObjectEntity.setNameLower(name);
-        contentType.ifPresent(catalogObjectEntity::setContentType);
-        contentType.ifPresent(catalogObjectEntity::setContentTypeLower);
-        catalogObjectRepository.save(catalogObjectEntity);
-        return new CatalogObjectMetadata(catalogObjectEntity);
+        List<CatalogObjectRevisionEntity> updateRevisionEntities = new ArrayList<>();
+        List<CatalogObjectEntity> updateCatalogObjectEntities = new ArrayList<>();
+
+        bucketNameObjectName.keySet().forEach(bucketName -> {
+            findBucketByNameAndCheck(bucketName);// TODO je pense pas necessaire car en dessous ca check aussi
```
fviale, line 367: if there is a TODO, it should be in english

src/main/java/org/ow2/proactive/catalog/service/CatalogObjectService.java
```diff
@@ -311,7 +311,7 @@ public CatalogObjectMetadata createCatalogObject(String bucketName, String name,
                                                                               catalogObjectEntity,
                                                                               metadataList,
                                                                               false);
-        return new CatalogObjectMetadata(result);
+        return new CatalogObjectMetadata(catalogObjectRevisionRepository.save(result));
```
fviale, line 314: buildCatalogObjectRevisionEntity already saves the new revision into the database, what is the need to save it again here ?

src/main/java/org/ow2/proactive/catalog/service/CatalogObjectService.java
AlexFalessi: I actually forgot to remove this. Thanks

src/main/java/org/ow2/proactive/catalog/service/CatalogObjectService.java
AlexFalessi: We save the catalog object entity + the revision entity.
There is no cascade.

fviale: last line of buildCatalogObjectRevisionEntity:
`return catalogObjectRevisionRepository.save(catalogObjectRevisionEntity);`

AlexFalessi: @fviale Ok you are confused because I modified the method. I removed the save in DB from there so that we can build multiple revision entities and update them in bulk after.

This method wasn't call so much so it wasn't a big change.

fviale (approved):

PR #343, opened 2025-03-21
add-tenant-filtering

oanaBiancaSchiopu:


src/main/java/org/ow2/proactive/catalog/rest/controller/BucketController.java
```diff
@@ -109,12 +109,27 @@ public BucketMetadata create(
             throws NotAuthenticatedException, AccessDeniedException {
         String initiator = ANONYMOUS;
         if (sessionIdRequired) {
-            restApiAccessService.checkAccessBySessionIdForOwnerOrGroupAndThrowIfDeclined(sessionId, ownerName);
+            AuthenticatedUser authenticatedUser = restApiAccessService.getUserFromSessionId(sessionId);
+            String tenant = authenticatedUser.getTenant();
```
fviale, line 113: this code is a bit strange

src/main/java/org/ow2/proactive/catalog/rest/controller/BucketController.java
```diff
@@ -109,12 +109,27 @@ public BucketMetadata create(
...
+            }
             initiator = restApiAccessService.getUserFromSessionId(sessionId).getName();
         }
         try {
-            BucketMetadata bucketMetadata = bucketService.createBucket(bucketName, ownerName);
-            log.info(ACTION + initiator + " created new bucket " + bucketName + " with owner " + ownerName);
+            String tenant = "";
+            if (sessionIdRequired) {
+                AuthenticatedUser authenticatedUser = restApiAccessService.getUserFromSessionId(sessionId);
+                tenant = authenticatedUser.getTenant();
```
fviale, line 128: tenant = authenticatedUser.getTenant() may return null, while you initialize tenant to empty string ...

fviale:
@oanaBiancaSchiopu 
After checking the PR, you use the configuration property  pa.catalog.tenant.filtering at the top-level (i.e. by setting tenant to null when set to false).
I don't think this is the proper way to do it. For example it implies that a bucket created will have no tenant information, and if the property is set to true later, it will not be possible to filter on created buckets in the meantime.

It is better to use the property at lower level

